PR #2, opened 2024-08-28
feat: add a kubeflow pipeline for SDG

tumido:
Pipeline accepts following parameters:

- Taxonomy git repo and branch or PR number.
   - If you specify a PR number, it runs SDG on diff `main..<PR_number>`.
   - If you specify a branch it does the same except for a local (to the repo) branch.  `main..<branch>`
   - If you don't specify either, it will run against the whole repo, as if there's no history of SDG.
- Inference model name, endpoint, token, number of instructions

Currently the pipeline fails at the SDG step since there's no diff to generate.

Pipeline produces 2 artifacts that are saved to S3:

- Clone of taxonomy repo (created in the first step of the pipeline, used by the second step)
- SDG generated dataset (created by the second step)

Pipeline uses:
- `ubi/toolbox` for `git clone` of taxonomy
- a selfmade container (UBI based) used as runtime for KFP python components. The overhead is minimal, see the `Containerfile`.

See an experiment here: https://rhods-dashboard-redhat-ods-applications.apps.ocp-beta-test.nerc.mghpcc.org/pipelines/tumido/pipelineRun/view/af3e108b-0985-4644-85d7-36db4afc8353

image

image

image

image

image


MichaelClifford:
Could you add an additional component like `update_taxonomy()` after `git-clone-op()` that would update the taxonomy based on user input and if no input is provided, make some default change to the repo. similar to below, just so it has something to generate data from?  This should overcome the failure at step 2.  https://github.com/redhat-et/ilab-on-ocp/blob/5686fdb3921850380c377e3d82b72ef0c5a62c0a/SDG/Containerfile#L11-L12

tumido:
I've managed to overcome that by checking out either a PR or a custom branch from the source repo. This way we do not have to hack things around. 🙂  If you don't specify the branch or PR param, you will get SDG for `empty` magic word - the whole repo.

tumido:
Updated. I had to use K8S CM/Secret instead of passing the token directly (as I should have done anyways). Now the pipeline doesn't truncate the token anymore and can talk to the inference endpoint.

I also had to fix `Permission denied` on HF cache by setting `HF_HOME` in the container image env.

Now the pipeline passes:
image

image


image

PR #3, opened 2024-08-28
initial single node training work

MichaelClifford:
This PR contains the initial  Containerfile, deployment yaml and training code needed to run a multi-gpu (but single node) training with instructlab on OpenShift.


training/Containerfile
```diff
@@ -0,0 +1,14 @@
+FROM nvcr.io/nvidia/cuda:12.4.1-devel-ubi9
+RUN dnf install -y python3.11 python3.11-devel git python3-pip make automake gcc gcc-c++
```
cooktheryan, line 2: @sallyom do we still need to a clean all here?

sallyom: yea, i believe we should add `&& dnf clean all` to that RUN cmd

training/run.py
```diff
@@ -0,0 +1,41 @@
...
+    data_path = "training/sample-data/train_all_pruned_SDG.jsonl",
+    ckpt_output_dir = "data/saved_checkpoints",
+    data_output_dir = "data/outputs",
+
+    # define model-trianing parameters
+    max_seq_len = 4096,
+    max_batch_len = 20000,
+    num_epochs = 2,
+    effective_batch_size = 3840,
+    save_samples = 250000,
```
JamesKunstle, line 22: you can set this to 0 if you want to disable save_samples checkpointing.

PR #4, opened 2024-08-28
Add single node pytorchjob yaml

Shreyanand:
This PR adds a single node pytorch job for the single node deployment created in this [PR](https://github.com/redhat-et/ilab-on-ocp/pull/3). The pytorch job can then be updated to add more worker nodes along with the master nodes. 

One caveat is that the entire [run.py](https://github.com/MichaelClifford/ilab-on-ocp/blob/train/training/run.py) script is in the yaml here. Maybe there is a better way to structure this. 

The pytorch job assumes the kubeflow training operator installed on the cluster which was done using the command:
`oc apply -k "github.com/kubeflow/training-operator.git/manifests/overlays/standalone?ref=v1.7.0"`

### DIFF
--- a/training/ilab-pytorch-job.yaml
+++ b/training/ilab-pytorch-job.yaml
@@ -1,8 +1,8 @@
 apiVersion: kubeflow.org/v1
 kind: PyTorchJob
 metadata:
-  name: ilab-test4
-  namespace: default
+  name: <TEST-NAME>
+  namespace: <NAMESPACE>
 spec:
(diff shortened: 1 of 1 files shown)


MichaelClifford:
Instead of having the run.py in the yaml, let's try and update the run.py along with the containerfile to work with the pytorchjob.

Shreyanand:
@MichaelClifford This works with the current `run.py`.

MichaelClifford (approved):
LGTM

PR #6, opened 2024-08-30
MT_Bench evaluation workflows

sallyom:
This PR adds 2 pipelines MT_Bench pipeline

@tumido please review

TODO in followup:
* create `launch vLLM with candidate model` task
* select best score from 10 candidates
* final evaluation

### DIFF
--- a/eval/Containerfile
+++ b/eval/Containerfile
@@ -4,22 +4,15 @@ RUN dnf install -y python3.11 python3.11-devel git python3-pip make automake gcc
     && dnf clean all \
     && rm -rf /var/cache/*dnf*
 
-WORKDIR /app
-
-COPY eval/requirements.txt .
 # TODO update to instructlab/eval after https://github.com/instructlab/eval/pull/128 or equivalent merges
-RUN git clone -b add-api-key https://github.com/sallyom/eval.git \
-    && rm -rf /usr/bin/python3 && ln -s /usr/bin/python3.11 /usr/bin/python3 \
-    && python3.11 -m ensurepip \
-    && python3.11 -m pip install --no-cache-dir -r ./eval/requirements.txt \
-    && python3.11 -m pip install --no-cache-dir -e ./eval \
-    && python3.11 -m pip install -r requirements.txt
+RUN python3.11 -m ensurepip \
+    && python3.11 -m pip install --no-cache-dir git+https://github.com/sallyom/eval@add-api-key \
+    && rm -rf /usr/bin/python3 && ln -s /usr/bin/python3.11 /usr/bin/python3
 
+ENV HF_HOME=/tmp
+ENV HOME=/tmp
+
+# required for mt_bench
 ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib"
-ENV HOME=/app
-
-# TODO: this
-RUN chmod -R 0777 /app
-USER 1001
 
 ENTRYPOINT ["python3.11"]
--- a/eval/mmlu_pipeline.py
+++ b/eval/mmlu_pipeline.py
@@ -1,6 +1,6 @@
 # type: ignore
 # pylint: disable=no-value-for-parameter,import-outside-toplevel,import-error
-from kfp import dsl, compiler
+from kfp import dsl, compiler, kubernetes
 from kfp.dsl import component, pipeline, Input, Output, Artifact, Model, importer
 
 EVAL_IMAGE = "quay.io/sallyom/instructlab-ocp:eval"
@@ -106,7 +106,8 @@ def mmlu_pipeline(
         mmlu_output=run_mmlu_task.output,
     )
 
-    #kubernetes.set_image_pull_policy(run_mmlu_task, "Always")
+    # TODO: remove this once images are set
+    kubernetes.set_image_pull_policy(run_mmlu_task, "Always")
     run_mmlu_task.set_accelerator_type('nvidia.com/gpu')
     run_mmlu_task.set_accelerator_limit(1)
 
--- a/eval/mt_bench_pipeline.py
+++ b/eval/mt_bench_pipeline.py
@@ -1,0 +1,185 @@
+# type: ignore
+# pylint: disable=no-value-for-parameter,import-outside-toplevel,import-error
+from typing import NamedTuple
+from kfp import dsl, compiler, kubernetes
+from kfp.dsl import component, pipeline, Input, Output, Artifact, Model, importer
+from kfp.kubernetes import use_config_map_as_env, use_secret_as_env
+
+EVAL_IMAGE = "quay.io/sallyom/instructlab-ocp:eval"
+TOOLBOX_IMAGE = "registry.access.redhat.com/ubi9/toolbox"
+CANDIDATE_S3_URI = "s3://sallyom-eval-e58df6b0-606b-4749-96a5-a105657cb068/models/instructlab/granite-7b-lab"
+K8S_NAME = "kfp-model-server"
+
+# TODO: This is a placeholder for launching vLLM with candidate model
+@component(base_image=EVAL_IMAGE)
+def serve_candidate_model(
+    candidate_model: Input[Model]
+) -> NamedTuple('outputs', candidate_model_name=str, candidate_server_url=str):
+    import os
+
+    candidate_model_name = os.path.basename(os.path.normpath(candidate_model.path))
+    print(f"(real)Candidate Model Name: {candidate_model_name}")
+    # TODO: REMOVE after implement vLLM with candidate model
+    # Fake model-name for now, with judge_model_name
+    candidate_model_name = os.getenv("JUDGE_MODEL_NAME") 
+    print(f"(fake)Candidate Model Name: {candidate_model_name}")
+
+    # TODO: Replace this with the candidate model endpoint
+    # Launch_server call with https://github.com/instructlab/instructlab/blob/main/src/instructlab/model/evaluate.py#L301C1-L327C1
+    candidate_server_url = os.getenv("JUDGE_MODEL_ENDPOINT")
+    outputs = NamedTuple('outputs', candidate_model_name=str, candidate_server_url=str)
+    return outputs(candidate_model_name, candidate_server_url)
+
+@component(base_image=EVAL_IMAGE)
+def run_mt_bench(
+    candidate_server_url: str,
+    candidate_model_name: str,
+    mt_bench_output: Output[Artifact],
+    device: str,
+    max_workers: str,
+    merge_system_user_message: bool,
+):
+    import json
+    import torch
+    import os
+    #from instructlab.eval.mt_bench import MTBenchEvaluator
+    from instructlab.eval import mt_bench_answers, mt_bench_judgment
+
+    os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
+
+    gpu_available = torch.cuda.is_available()
+    gpu_name = torch.cuda.get_device_name(torch.cuda.current_device()) if gpu_available else "No GPU available"
+
+    print(f"GPU Available: {gpu_available}, Using: {gpu_name}")
+
+    effective_device = device if device is not None else ("cuda" if gpu_available else "cpu")
+
+    print(f"Running on device: {effective_device}")
+
+    judge_model_name = os.getenv("JUDGE_MODEL_NAME")
+    judge_endpoint = os.getenv("JUDGE_MODEL_ENDPOINT")
+    print(f"Candidate Server URL: {candidate_server_url}")
+    print(f"Candidate Model Name: {candidate_model_name}")
+
+    if max_workers == "auto":
+        try:
+            usable_cpu_count = len(os.sched_getaffinity(0)) // 2
+        except AttributeError:
+            usable_cpu_count = multiprocessing.cpu_count() // 2
+        max_workers = usable_cpu_count
+
+    # TODO: Using evaluator results in connection errors, need to determine why.
+    #       For now, using mt_bench_answers.generate_answers & mt_bench_judgment.generate_judgment
+    #evaluator = MTBenchEvaluator(
+    #    model_name=candidate_model_name,
+    #    judge_model_name=judge_model_name,
+    #    max_workers=max_workers,
+    #    merge_system_user_message=merge_system_user_message
+    #)
+
+    judge_api_key = os.getenv("JUDGE_API_KEY", "")
+
+    print("Generating answers...")
+    # TODO: launch vLLM with candidate-model
+    # Faking it by using judge_endpoint for generate, so need to pass the api_key & candidate_server_url is actually judge_server_url for now
+    # Need to add a launch_server task with https://github.com/instructlab/instructlab/blob/main/src/instructlab/model/evaluate.py#L301C1-L327C1
+    # with local vLLM, won't need to pass an api_key here
+    #
+    # See note above about issue with evaluator
+    #evaluator.gen_answers(server_url=candidate_server_url, api_key=judge_api_key)
+    mt_bench_answers.generate_answers(
+        model_name=candidate_model_name,
+        model_api_base=candidate_server_url,
+        api_key=judge_api_key,
+        output_dir="/tmp/eval_output",
+        max_workers=max_workers
+    )
+
+    print("Judging answers...")
+    # See note above about issue with evaluator
+    #overall_score, qa_pairs, turn_scores = evaluator.judge_answers(server_url=judge_endpoint, api_key=judge_api_key)
+    overall_score, qa_pairs, turn_scores, error_rate = mt_bench_judgment.generate_judgment(
+        model_name = candidate_model_name,
+        judge_model_name=judge_model_name,
+        model_api_base=judge_endpoint,
+        api_key=judge_api_key,
+        output_dir="/tmp/eval_output",
+        max_workers=max_workers,
+        merge_system_user_message=merge_system_user_message
+    )
+
+    mt_bench_data = {
+        "report_title": "SKILLS EVALUATION REPORT",
+        "model": candidate_model_name,
+        "judge_model": judge_model_name,
+        "overall_score": overall_score,
+        "turn_scores": turn_scores,
+        "qa_scores": qa_pairs,
+        "error_rate": error_rate,
+    }
+
+    with open(mt_bench_output.path, 'w') as f:
+        json.dump(mt_bench_data, f, indent=4)
+
+@component(base_image=TOOLBOX_IMAGE)
+def load_mt_bench_results(mt_bench_output: Input[Artifact]) -> dict:
+    import json
+
+    with open(mt_bench_output.path, 'r') as file:
+        mt_bench_data = json.load(file)
+    print("Loaded MT_Bench Data:")
+    print(json.dumps(mt_bench_data, indent=4))
+
+    return mt_bench_data
+
+@pipeline(
+    display_name="MT_BENCH Evaluation Pipeline",
+    name="mt_bench_eval",
+    description="Pipeline to run the MT_BENCH evaluation",
+)
+
+def mt_bench_pipeline(
+    # generate_answers,judgment uses a magic word for its mt_bench evaluator  - `auto`
+    # with `auto`, number of gpus allocated for serving is calculated based on environment
+    # https://github.com/instructlab/eval/blob/main/src/instructlab/eval/mt_bench.py#L36 
+    max_workers: str = "auto",
+    merge_system_user_message: bool = False,
+    device: str = None,
+
+):
+    import_candidate_model_task = importer(
+        artifact_uri=CANDIDATE_S3_URI,
+        artifact_class=Model,
+        # TODO: True/False?
+        reimport=False,
+    )
+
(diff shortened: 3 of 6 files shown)


eval/Containerfile
```diff
@@ -0,0 +1,21 @@
...
+
+RUN export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib"
+ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib"
+ENV HOME=/instructlab
+
+RUN chmod -R 777 /instructlab
+USER 1001
+COPY eval/eval-scripts .
+
+ENTRYPOINT ["/bin/bash"]
```
tumido, line 21: Set this to

```
CMD ["python"]
```

See my PR for details: https://github.com/redhat-et/ilab-on-ocp/pull/2/files#diff-602239348cc944737a4fc15b5d1b8509bf60ddacf0152b5027e4db16916312acR4

Then you can turn your `@dsl.container_components` into `@dsl.components` and run your python directly - no need for shell -> python -> script overhead.

eval/requirements.txt
```diff
@@ -0,0 +1,14 @@
...
+psutil
+lm-eval>=0.4.3
+boto3
+requests
+torch
+transformers
+accelerate
+pandas
+pandas-stubs
+kfp[kubernetes]
```
tumido, line 14: The pipeline component (the container image) needs to contain `kfp[kubernetes]`?

The rest seems to be the same as https://github.com/instructlab/eval/blob/main/requirements.txt, except `boto3` and `requests`.

The dependencies that area already stated in https://github.com/instructlab/eval/blob/main/requirements.txt has been previously already installed when you ran

```
pip install --no-cache-dir git+https://github.com/instructlab/eval
```

Maybe we can reduce this to:

```suggestion
boto3
requests
```

eval/Containerfile
```diff
@@ -0,0 +1,21 @@
...
+
+RUN dnf install -y python3.11 python3.11-devel git python3-pip make automake gcc gcc-c++ \
+    && dnf clean all \
+    && rm -rf /var/cache/*dnf*
+
+WORKDIR /instructlab
+
+COPY eval/requirements.txt .
+RUN python3.11 -m ensurepip
+RUN python3.11 -m pip install --no-cache-dir git+https://github.com/instructlab/eval && python3.11 -m pip install -r requirements.txt
```
tumido, line 11:
```suggestion
RUN python3.11 -m ensurepip && python3.11 -m pip install --no-cache-dir git+https://github.com/instructlab/eval && python3.11 -m pip install --no-cache-dir  -r requirements.txt
```

eval/Containerfile
```diff
@@ -0,0 +1,21 @@
...
+
+RUN export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib"
```
tumido, line 13: Why this layer? `RUN` doesn't persist content to next images. Therefore the `export` command has no effect on the resulting image.

eval/Containerfile
```diff
@@ -0,0 +1,21 @@
...
+ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib"
+ENV HOME=/instructlab
+
+RUN chmod -R 777 /instructlab
```
tumido, line 17: This doesn't look very safe

eval/eval-scripts/mt_bench.py
```diff
@@ -0,0 +1,149 @@
...
+            print(f"Downloaded {key} to {local_file_path}")
+
+        if os.listdir(model_local_path):
+            print(f"Model files found at {model_local_path}.")
+        else:
+            print(f"Model files not found at {model_local_path}. Check the S3 path and permissions.")
+
+    except Exception as e:
+        print(f"Error downloading model from S3: {e}")
+        raise
```
tumido, line 83: Same as above:

Can't you use here `dsl.importer` instead? It will be safer + less code to maintain + always using tracked artifacts.

https://www.kubeflow.org/docs/components/pipelines/user-guides/components/importer-component/

Maintaining 2 times 40 lines of S3 downloader code seems like a tech debt.

eval/eval-scripts/mmlu.py
```diff
@@ -0,0 +1,138 @@
...
+            # Verify that the expected files are downloaded
+            existing_files = os.listdir(local_model_path)
+            if all(file in existing_files for file in required_files):
+                logger.debug(f"Required model files found at {local_model_path}.")
+            else:
+                logger.debug(f"Model files not found at {local_model_path}. Expected files: {required_files}. Check the S3 path and permissions.")
+
+        except Exception as e:
+            logger.debug(f"Error downloading model from S3: {e}")
+            raise
```
tumido, line 82: Can't you use here `dsl.importer` instead? It will be safer + less code to maintain + always using tracked artifacts.

https://www.kubeflow.org/docs/components/pipelines/user-guides/components/importer-component/

eval/eval-scripts/mmlu.py
```diff
@@ -0,0 +1,138 @@
...
+    logger.debug("MMLU eval phase -- Clearing PyTorch cache")
+    torch.cuda.empty_cache()
+
+    with open(output_file, 'w') as f:
+        f.write(f"Overall MMLU Score: {overall_score}")
+    logger.debug(f"Saved evaluation results to {output_file}")
+
+    try:
+        output_prefix = '/'.join(output_s3_path.split('/')[3:])
+        s3_client.upload_file(output_file, bucket_name, f"{output_prefix}/{os.path.basename(output_file)}")
```
tumido, line 120: You can produce `dsl.Output` instead of uploading to S3 directly.

If you need output artifact path - local to the pod (you can write to it as if it was a file/folder), you can use:

```python
a: dsl.Output

with open(a.path, "w") as f:
...
```

eval/mt_bench_pipeline.py
```diff
@@ -0,0 +1,129 @@
...
+
+bucket_name = eval_script_s3_path.split('/')[2]
+key = '/'.join(eval_script_s3_path.split('/')[3:])
+
+try:
+    s3_client.download_file(bucket_name, key, local_path)
+    print(f'Downloaded {{key}} to {{local_path}}')
+except Exception as e:
+    print(f'Error downloading mt_bench script from S3: {{e}}')
+    raise
```
tumido, line 69: This is the 3rd time you're downloading from S3, using yet another code snippet - 3rd way how to download a thing... This time in a string. This is definitely a danger.

tumido (changes requested):
PTAL at #2 and turn the container components into `dsl.components` it will be a lot less of an overhead + much easier to maintain.

sallyom:
@tumido this wasn't ready for review :) but thank you anyways ❤️

eval/eval-scripts/mt_bench.py
sallyom: why yes I can

eval/requirements.txt
sallyom: I dunno why but these were required to make the workflow proceed - I'm reworking everything so will look at that with new versions

eval/Containerfile
sallyom: you have to trust that users can handle this amount of power 🤣

eval/Containerfile
sallyom: I wonder if I use `CMD ["python3.11"` instead of `ENTRYPOINT ["3.11"]` dsl will indeed run 3.11 and not 3.9. From my empirical data, dsl runs 3.9 in this container not 3.11. I had to explicity set 3.11 in the generated yaml 🤷‍♀️ but yes, I've updated to use component rather than the container->shell->python

eval/mt_bench_pipeline.py
sallyom: > WIP: Don't Review!! Don't try to run (probably won't run anyways) :)

We don't need S3 at all, I replaced with importer and Artifacts

eval/eval-scripts/mmlu.py
sallyom: yup

> WIP: Don't Review!! Don't try to run (probably won't run anyways) :)

eval/eval-scripts/mmlu.py
sallyom: yup

> WIP: Don't Review!! Don't try to run (probably won't run anyways) :)

eval/Containerfile
sallyom: https://github.com/redhat-et/ilab-on-ocp/pull/6/commits/2c2bd58ae03137ed6d3b6998246fd8f67825110d

sallyom:
Screenshot 2024-09-10 at 11 31 23 PM
mmlu

example mmlu output:
```
$ cat ~/Downloads/mmlu_output_file
# KNOWLEDGE EVALUATION REPORT
## MODEL
granite-7b-lab
### AVERAGE:
0.42 (across 3)
### SCORES:
mmlu_abstract_algebra - 0.3
mmlu_anatomy - 0.44
mmlu_astronomy - 0.53
```

eval/requirements.txt
```diff
@@ -0,0 +1 @@
...
```
tumido, line 1: I guess this ends up in the `Containerfile` - in the image, right? Why do we need the DSL to be there?

eval/mmlu_pipeline.py
```diff
@@ -0,0 +1,116 @@
...
+        f.write(f"# KNOWLEDGE EVALUATION REPORT\n")
+        f.write("\n## MODEL\n")
+        f.write(f"{model_name}\n")
+        f.write(f"\n### AVERAGE:\n")
+        f.write(f"{round(mmlu_score, 2)} (across {len(individual_scores)})\n")
+
+        f.write(f"### SCORES:\n")
+        for task, score in individual_scores.items():
+            s = round(score["score"], 2)
+            f.write(f"{task} - {s}\n")
```
tumido, line 60: Just a thought, can we write it as JSON? Or CSV? It would be much easier to process as output if we need to.

eval/mt_bench_pipeline.py
```diff
@@ -0,0 +1,145 @@
...
+    from instructlab.eval.mt_bench import MTBenchEvaluator
+
+    os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
+
+    gpu_available = torch.cuda.is_available()
+    gpu_name = torch.cuda.get_device_name(torch.cuda.current_device()) if gpu_available else "No GPU available"
+
+    print(f"GPU Available: {gpu_available}, Using: {gpu_name}")
+
+    effective_device = device if device != "None" else ("cuda" if gpu_available else "cpu")
```
tumido, line 47: Where the `device` variable comes from? I can't see it, am I blind?

Also, you can use `None` type, instead of `"None"` string. That would be better I think.

eval/mt_bench_pipeline.py
```diff
@@ -0,0 +1,145 @@
...
+
+    print(f"Running on device: {effective_device}")
+
+    model_name = os.getenv("CANDIDATE_MODEL_NAME")
+
+    if max_workers == "auto":
```
tumido, line 53: Can this be optional param instead of a magic word?

```suggestion
    if not max_workers:
```

tumido (changes requested):
Just a few nits, here and there, otherwise LGTM. Great job on this! 👍 🙌

eval/Containerfile
```diff
@@ -0,0 +1,24 @@
...
+
+COPY eval/requirements.txt .
+RUN git clone https://github.com/instructlab/eval.git \
+    && rm -rf /usr/bin/python3 && ln -s /usr/bin/python3.11 /usr/bin/python3 \
+    && python3.11 -m ensurepip \
+    && python3.11 -m pip install --no-cache-dir -r ./eval/requirements.txt \
+    && python3.11 -m pip install --no-cache-dir -e ./eval \
+    && python3.11 -m pip install -r requirements.txt
+
+ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib"
```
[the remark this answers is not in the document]
sallyom: I figured this is how python knows where cuda pkgs are but I haven't looked much further than that

eval/mt_bench_pipeline.py
```diff
@@ -0,0 +1,145 @@
...
+from kfp.kubernetes import use_config_map_as_env, use_secret_as_env
+
+EVAL_IMAGE = "quay.io/sallyom/instructlab-ocp:eval"
+TOOLBOX_IMAGE = "registry.access.redhat.com/ubi9/toolbox"
+CANDIDATE_S3_URI = "s3://sallyom-eval-e58df6b0-606b-4749-96a5-a105657cb068/models/instructlab/granite-7b-lab"
+K8S_NAME = "kfp-model-server"
+
+# TODO: This is a placeholder for launching vLLM with candidate model
+@component(base_image=EVAL_IMAGE)
+def serve_candidate_model(
```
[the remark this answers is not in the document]
sallyom: no the judge model is the mistral, centrally served - the candidate model will be served with an internal vLLM, like here: https://github.com/instructlab/instructlab/blob/main/src/instructlab/model/evaluate.py#L299-#L341

tumido (changes requested):

eval/mmlu_pipeline.py
sallyom: updated!

eval/Containerfile
```diff
@@ -0,0 +1,25 @@
...
+WORKDIR /app
+
+COPY eval/requirements.txt .
+# TODO update to instructlab/eval after https://github.com/instructlab/eval/pull/128 or equivalent merges
+RUN git clone -b add-api-key https://github.com/sallyom/eval.git \
+    && rm -rf /usr/bin/python3 && ln -s /usr/bin/python3.11 /usr/bin/python3 \
+    && python3.11 -m ensurepip \
+    && python3.11 -m pip install --no-cache-dir -r ./eval/requirements.txt \
+    && python3.11 -m pip install --no-cache-dir -e ./eval \
+    && python3.11 -m pip install -r requirements.txt
```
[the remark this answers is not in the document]
sallyom: I have to clone the repo to get the fix for adding api_key (for now, until it merges) - also I needed to install the requirements.txt from the repo, I was getting "module not found" when only installing the python package (make sense?) - either that or I needed to add the list of requirements in the requirements.txt in this repo.

sallyom:
@MichaelClifford @tumido Look, mt_bench output, yay 

 mt_bench_out

```
   "report_title": "SKILLS EVALUATION REPORT",
    "model": "mistral-7b-instruct-v02",
    "judge_model": "mistral-7b-instruct-v02",
    "overall_score": 8.803225806451612,
    "turn_scores": [
        8.871794871794872,
        8.733766233766234
    ],
    "qa_scores": [
        {
            "question_id": "107",
            "score": 10.0,
            "category": "reasoning",
            "question": [
                "A is the father of B. B is the father of C. What is the relationship between A and C?",
                "Building on the previous question, if C is the son of D, D is the father of E, E is the son of X, and X is the father of Y, and Y is the father of Z, what's the relationship between A and Z in terms of generations and also the familial relationship in words?"
            ],
            "answer": [
                {
                    "index": 0,
                    "turns": [
                        " A is the grandfather of C.\n\nHere's an explanation:\n\n1. We know that A is the father of B.\n2. We also know that B is the father of C.\n3. Therefore, A is the grandparent of C, specifically the grandfather.\n\nSo, A's relationship to C is that of a grandfather.",
                        " A is the grandfather of Z, and they are three generations apart.\n\nHere's an explanation:\n\n1. We know that A is the father of B.\n2. We also know that B is the father of C.\n3. C is the father of D.\n4. D is the father of E.\n5. E is the father of X.\n6. X is the father of Y.\n7. Y is the father of Z.\n\nSo, A is the grandfather of Z, and there are three generations between them. In words, A is the grandfather of Z, meaning Z is the grandchild of A."
                    ]
                }
            ]
        },
---------
```

eval/requirements.txt
sallyom: I think I remember seeing a "module not found" error without it - but probably you are right, since you didn't need it for the sdg workflow - I've removed it.

eval/mt_bench_pipeline.py
sallyom: updated.

eval/mt_bench_pipeline.py
sallyom: it's a magical word in instructlab/eval: https://github.com/instructlab/eval/blob/main/src/instructlab/eval/mt_bench.py#L36

eval/requirements.txt
tumido: I'm confused. Did you forget to push the latest changes or you didn't remove it?

eval/mt_bench_pipeline.py
tumido: Can you please comment about it in the code then?

I did something similar here:

https://github.com/redhat-et/ilab-on-ocp/blob/650077f8b215bd87292a3bdd0315a4dd1413eabd/sdg/components.py#L52-L54

eval/mt_bench_pipeline.py
tumido: I'm confused. Did you forget to push the latest changes or you didn't update it?

eval/Containerfile
```diff
@@ -6,7 +6,6 @@ RUN dnf install -y python3.11 python3.11-devel git python3-pip make automake gcc
 
 WORKDIR /app
 
-COPY eval/requirements.txt .
```
tumido, line 9: Please update the `RUN` command below as well. `eval/requirements.txt` doesn't exist anymore.

Can you please apply also the other comments I had here?
https://github.com/redhat-et/ilab-on-ocp/pull/6#discussion_r1750146938


Since you're pulling from your fork it should ideally look like this:

```
RUN python3.11 -m ensurepip \
    && python3.11 -m pip install --no-cache-dir git+https://github.com/sallyom/eval@add-api-key
```

That's it. No repo cloning (the repo ends up in the image with whole git history, that is not necessary at all), no extra `requirements.txt`s.

eval/Containerfile
```diff
@@ -6,7 +6,6 @@ RUN dnf install -y python3.11 python3.11-devel git python3-pip make automake gcc
 
 WORKDIR /app
```
tumido, line 7: Can you please also address the:

```suggestion
WORKDIR /app
```

and later

```
ENV HOME=/app

# TODO: this
RUN chmod -R 0777 /app
```

Why are you getting permission issues and why does it have to be `/app` folder?

Would something like `HF_HOME=/tmp` save you all the headaches? Or is the issue somewhere else?

tumido (changes requested):

eval/requirements.txt
sallyom: oh I thought I removed it - 👀 maybe didn't push that.....

eval/mt_bench_pipeline.py
sallyom: must've not pushed... 👀

eval/Containerfile
sallyom: I was getting permission errors & wanted to quickly get past them, then never went back to it, lemme test to make sure all is good with /tmp - and might want to update this, too: https://github.com/redhat-et/ilab-on-ocp/blob/main/training/Containerfile#L14

eval/Containerfile
sallyom: but I'll test it again to be sure it's good w/out the extra requirements.txt

eval/requirements.txt
sallyom: updated, done

eval/mt_bench_pipeline.py
sallyom: updated, pushed, done

eval/mt_bench_pipeline.py
sallyom: updated with note, done

eval/Containerfile
```diff
@@ -0,0 +1,25 @@
+FROM nvcr.io/nvidia/cuda:12.4.1-devel-ubi9
+
+RUN dnf install -y python3.11 python3.11-devel git python3-pip make automake gcc gcc-c++ \
+    && dnf clean all \
+    && rm -rf /var/cache/*dnf*
+
+WORKDIR /app
+
+COPY eval/requirements.txt .
```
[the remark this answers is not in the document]
sallyom: you are correct - so, what happened was I wasn't in a venv locally when running and I was confused 😆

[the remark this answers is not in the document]
sallyom: removed

eval/Containerfile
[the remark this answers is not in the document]
sallyom: same, confused and was in the wrong venv - updated

eval/Containerfile
sallyom: all good :) (you already knew that) removed, updated!

eval/Containerfile
sallyom: updated to /tmp all set

sallyom:
ready for merge, will work for now - TODOs listed above & am working on those now

MichaelClifford (approved):
LGTM!

PR #9, opened 2024-08-30
Initial work for sdg on ocp

MichaelClifford:
Initial Containerfile, deployment.yaml and python script to run SDG on OCP. 

Admittedly this is a fairly rigid example, where I've updated the `Taxonomy/` repo  as part of the container build so that it will always start with a git diff for the SDG to use (SDG seems to require a git diff to pick out what data will be used for generation). Currently there is no real way for a user to update the taxonomy with this code, hopefully we can address that in a later PR when we determine how a user will actually interact with this whole pipeline. 

To run this code on OCP you must: 
* set OPENAI_API_KEY to your Authorization Token for your route 
* OPENAI_BASE_URL to your route address
* OPENAI_MODEL to the model name in the vllm model server.


cooktheryan:
@tumido

eval/deployment.yaml
```diff
@@ -0,0 +1,32 @@
...
+      containers:
+        - name: container
+          image: quay.io/michaelclifford/test-sdg:0.0.0
+          resources:
+          command:
+            - /bin/bash
+            - '-c'
+            - '--'
+          args:
+            - while true; do sleep 300; done
```
Shreyanand, line 25: should we add `python3.11 eval/run.py` here

Shreyanand (commented):
Looks good overall 👍

hemajv (commented):
Any reason these files are under the `/eval` folder and not the `/SDG` folder?

Else, everything else looks good to me 👍

eval/deployment.yaml
MichaelClifford: yeah, sure.

MichaelClifford:
@hemajv Good catch, lol. definitely in the wrong directory 😄

SDG/deployment.yaml
```diff
@@ -0,0 +1,32 @@
...
+      containers:
+        - name: container
+          image: quay.io/michaelclifford/test-sdg:0.0.0
+          resources:
+          command:
+            - /bin/bash
+            - '-c'
+            - '--'
+          args:
+            - while true; do sleep 300; done
```
JudeNiroshan, line 25:
```suggestion
            - python3.11 run.py; while true; do sleep 300; done
```

SDG/run.py
```diff
@@ -0,0 +1,23 @@
...
+
+if not os.getenv("OPENAI_API_KEY"):
+    os.environ["OPENAI_API_KEY"] = "Not_Needed"
+if not os.getenv("OPENAI_BASE_URL"):
+    os.environ["OPENAI_BASE_URL"]  = "http://10.88.0.1:8001/v1"
+if not os.getenv("OPENAI_MODEL"):
+     os.environ["OPENAI_MODEL"] = "Not_Needed"
+
+client = openai.OpenAI()
+openai.default_headers = {"Authorization": f'Bearer:{os.environ["OPENAI_API_KEY"]}'}
```
JudeNiroshan, line 13: Should we be more specific about this? something like `MODEL_SERVER_TOKEN`

JudeNiroshan (approved):

SDG/run.py
hemajv: +1, I think that would provide more clarity

tumido:
Do we need this when we have functional #2 ?🙂

MichaelClifford:
@tumido yes, lets close this PR in favor of #2. It looks like all the relevant bits have been incorporated over there 😄

PR #13, opened 2024-09-04
kubernetes items

cooktheryan:
We could set this up to run during slow times. We will need to further tune this and allow for it to be done by users before running the pipeline


kubernetes_yaml/model_dowloader/container_file/download_hf_model.py
```diff
@@ -86,6 +89,13 @@ def save_model_to_s3(local_model_path: Union[str, Path], s3_model_path: str):
 
     return s3_path
 
+# Example usage with verbosity
+save_model_to_s3(
```
hemajv, line 93: We should probably just comment this out, since it will be invoked automatically when the entire script is run as-is

kubernetes_yaml/model_dowloader/job/download_job.yaml
```diff
@@ -0,0 +1,40 @@
...
+          valueFrom:
+            secretKeyRef:
+              name: model-store
+              key: AWS_ACCESS_KEY_ID
+        - name: AWS_SECRET_KEY
+          valueFrom:
+            secretKeyRef:
+              name: model-store
+              key: AWS_SECRET_ACCESS_KEY
+        - name: S3_ENDPOINT
```
hemajv, line 34: we should also add another env var for the `S3_FOLDER` which is being used here: https://github.com/redhat-et/ilab-on-ocp/blob/downloader/kubernetes_yaml/model_dowloader/container_file/download_hf_model.py#L69

hemajv (changes requested):
thanks for the changes @cooktheryan, left a few comments ptal

hemajv (approved):
thanks for the changes @cooktheryan!
/lgtm 🚢

PR #10, opened 2024-09-03
Add draft YAML for distributed training.

ilya-kolchinsky:

PR #14, opened 2024-09-06
Add kfp pipeline for running a pytorch job

Shreyanand:
This PR adds the kfp pipeline for launching the Instructlab train Pytorch job. It is based on the kfp pytorch launcher [component](https://github.com/kubeflow/pipelines/tree/master/components/kubeflow/pytorch-launcher). However that component is 3 years old and I had to make some changes for it to work. 

In the current state, the pipeline works- kfp workflow triggers a pytorch job. However it is not able to run multinode training yet. Fixes found in #10 should make this run completely. 

One caveat with the current launcher is that it does not have a way to run the pytorch job with just the master node. It has to have a worker node. That is not a limitation of a vanilla pytorch job. 

Putting it in draft for now but it can be tested if curious.

### DIFF
--- a/training/pipeline.py
+++ b/training/pipeline.py
@@ -1,0 +1,148 @@
+from typing import NamedTuple
+import kfp.dsl as dsl
+from kfp import components
+
+@dsl.component(base_image="python:slim")
+def create_worker_spec(worker_num: int = 0) -> NamedTuple(
+    "CreatWorkerSpec", [("worker_spec", dict)]):
+    """
+    Creates pytorch-job worker spec
+    """
+    from collections import namedtuple
+    worker = {}
+    if worker_num > 0:
+        worker = {
+            "replicas": worker_num,
+            "restartPolicy": "OnFailure",
+            "template": {
+                "metadata": {
+                    "annotations": {
+                        "sidecar.istio.io/inject": "false"
+                    }
+                },
+                "spec": {
+                    "containers": [
+                        {   "command": [
+                            '/bin/bash',
+                            '-c',
+                            '--'
+                            ],
+                            "args": [
+                            "python3.11 -u run.py"
+                            ],
+                            "image": "quay.io/michaelclifford/test-train:0.0.11",
+                            "name": "pytorch",
+                            "resources": {
+                                "requests": {
+                                    "memory": "8Gi",
+                                    "cpu": "2000m",
+                                    # Uncomment for GPU
+                                    "nvidia.com/gpu": 1,
+                                },
+                                "limits": {
+                                    "memory": "8Gi",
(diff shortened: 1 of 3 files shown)


cooktheryan:
@tumido do you want to review this in the current state and @MichaelClifford maybe Monday we pair this with Tom's PR and start generating the MVP run

training/pipeline.py
```diff
@@ -0,0 +1,148 @@
...
+                                "cpu": "2000m",
+                                # Uncomment for GPU
+                                "nvidia.com/gpu": 1,
+                            },
+                            "limits": {
+                                "memory": "8Gi",
+                                "cpu": "2000m",
+                                # Uncomment for GPU
+                                "nvidia.com/gpu": 1,
+                            },
```
tumido, line 111: Should we parametrize this?

tumido (approved):
Can we please remove the `create_worker_spec` component? It seems to be unnecessary here.

tumido (changes requested):
Wrong button 😄  I've meant to request changes... 😄

training/pipeline.py
```diff
@@ -0,0 +1,148 @@
...
+        ttl_seconds_after_finished=ttl_seconds_after_finished,
+        job_timeout_minutes=job_timeout_minutes,
+        delete_after_done=delete_after_done,
+        active_deadline_seconds=100,
+        backoff_limit=1
+    )
+
+
+if __name__ == "__main__":
+    import kfp.compiler as compiler
```
tumido, line 140: I don't think you need to nest the import here. This is against https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/import-outside-toplevel.html

training/pipeline.py
```diff
@@ -0,0 +1,148 @@
...
+
+    pipeline_file = "pipeline.yaml"
+    print(
+        f"Compiling pipeline as {pipeline_file}"
+    )
+    compiler.Compiler().compile(
+        ilab_train, pipeline_file
+    )
```
tumido, line 148: This is somewhat weirdly formatted. I think simple:


```suggestion
    pipeline_file = "pipeline.yaml"
    print(f"Compiling pipeline as {pipeline_file}")
    compiler.Compiler().compile(ilab_train, pipeline_file)
```

Would be fully compliant with PEP8.

tumido (changes requested):

leseb:
Please install https://docs.astral.sh/ruff/installation/ on your system and run `ruff format .` once https://github.com/redhat-et/ilab-on-ocp/pull/57 is merged. THEN rebase your PR, everything should go fine :).

Shreyanand:
Closing in favour of #231

PR #20, opened 2024-09-11
feat(kfp): add train stage

tumido:
Competition PR to #14 

This bring Training stage to the pipeline

This is how it looks when run with faked SDG (I didn't want to wait for the PyTorchJob so I made it timeout on the wait step):

image

What works:
- Scheduling PyTorchJob
- Waiting for it
- Consuming SDG output and passing it to PVC
- Downloading base model from HuggingFace and storing it in Artifact
- Populating a PVC with base model
- Creating a PVC for output collection
- Collecting outputs from PVC into artifacts (Model + other outputs)
- Mounting all PVC to PyTorchJob

TODO:
- ~Fix passing data to PytorchJob, PVC requirement may not be the best option here~
- Fix `kubectl wait` to allow timeout/namespace params, I was initially not successful here due to some KFP templating issues...
- See if we can use something better for PyTorchJob `metadata.name` until `dsl.PIPELINE_*` is available (https://github.com/kubeflow/pipelines/issues/10453)
- See if we can do better than the storage class enforced by KFP: https://github.com/kubeflow/pipelines/blob/1cded35cf5e93d8c8d32fefbddceb2eed8de9a0a/backend/src/v2/driver/driver.go#L1428-L1436
- Add params/args everywhere for more customization

### DIFF
--- a/training/Containerfile
+++ b/training/Containerfile
@@ -18,6 +18,14 @@ RUN export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-pack
     && python3.11 -m pip install packaging wheel setuptools-scm \
     && python3.11 -m pip install --no-cache-dir "instructlab-training[cuda]@git+https://github.com/instructlab/training.git"
 
+# We don't expect to keep Triton cache after pod is finished
+ENV TRITON_CACHE_DIR="/tmp"
+
+# Set HF_HOME to volatile location. We do not preserve cache etc over restarts with this container.
+# https://huggingface.co/docs/huggingface_hub/en/guides/manage-cache
+ENV HF_HOME="/tmp"
+ENV TRANSFORMERS_CACHE="/tmp"
+
 COPY run.py .
 
 USER 1001
--- a/training/Containerfile
+++ b/training/Containerfile
@@ -1,16 +1,24 @@
 FROM nvcr.io/nvidia/cuda:12.4.1-devel-ubi9
+
 RUN dnf install -y python3.11 python3.11-devel git python3-pip make automake gcc gcc-c++ \
     && dnf clean all \
     && rm -rf /var/cache/*dnf*
-WORKDIR /instructlab
-RUN python3.11 -m ensurepip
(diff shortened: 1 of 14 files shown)


training/components.py
```diff
@@ -0,0 +1,138 @@
...
+                          value: \\"{nnodes}\\"
+                        - name: NPROC_PER_NODE
+                          value: \\"{nprocPerNode}\\"
+                      resources:
+                        requests:
+                          memory: 8Gi
+                          cpu: 2
+                          "nvidia.com/gpu": {nprocPerNode}
+                        limits:
+                          memory: 8Gi
```
MichaelClifford, line 74: I think this memory limit is likely too low. The model needs to be loaded into memory which is at least 14 GB. Would it me possible to remove the  memory limit while we are developing the pipeline?

training/Containerfile
```diff
@@ -1,16 +1,32 @@
...
+RUN export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib" \
+    && python3.11 -m ensurepip \
+    && python3.11 -m pip install --no-cache-dir --upgrade pip \
+    && CMAKE_ARGS="-DLLAMA_CUDA=on" python3.11 -m pip install "instructlab-training@git+https://github.com/instructlab/training.git" \
+    && python3.11 -m pip install packaging wheel setuptools-scm \
+    && python3.11 -m pip install --no-cache-dir "instructlab-training[cuda]@git+https://github.com/instructlab/training.git"
+
+# We don't expect to keep Triton cache after pod is finished
+ENV TRITON_CACHE_DIR="/tmp"
+
```
MichaelClifford, line 23: There is an `[rank1]: PermissionError: [Errno 13] Permission denied: '/.cache'` error during training, as torch attempts to write to the .cache dir. We can fix it by setting the following env variable.  
```suggestion
XDG_CACHE_HOME=/tmp
```

Shreyanand (approved):
LGTM in the current state. The PR adds most of the sdg + training pipeline skeleton. We could merge and iterate over it.

Issue #19, opened 2024-09-11
Add preflight checks in the pipeline.py


JudeNiroshan:
Exit pipeline with more concise information when required input data is not available as expected.

- Model Server URL
- Model Server token
- Model name
- Taxonomy repository branch AND PR number

jgarciao:
In RHOAI 2.19 there is a task in the pipeline checking the connection with the teacher, judge, model registry, oci registry and availability of PytorchJob resource in the cluster. I think this covers most of what it was proposed here

![Image](https://github.com/user-attachments/assets/9cf227d8-a8e8-4612-9c99-fb57ef7d16fe)

PR #21, opened 2024-09-12
Add basic preflight checks as the first component in the pipeline

JudeNiroshan:
Address #19

### DIFF
--- a/sdg/components.py
+++ b/sdg/components.py
@@ -5,6 +5,26 @@ from kfp import dsl
 
 IMAGE = "quay.io/tcoufal/ilab-sdg:latest"
 
+@dsl.component(base_image=IMAGE)
+def preflight_check_op(
+    repo_branch: str,
+    repo_pr: Optional[int],
+):
+    from os import getenv
+
+    if (not repo_branch) and (repo_pr is None or repo_pr <= 0 ):
+        raise Exception("Both taxonomy repo branch and taxonomy pull request number cannot be empty")
+    api_key = getenv("api_key")
+    model = getenv("model")
+    endpoint = getenv("endpoint")
+
+    if not api_key:
+        raise Exception("Model Server Auth Key is missing in kfp-model-server secret")
+    if not model:
+        raise Exception("Model name is missing in kfp-model-server configMap")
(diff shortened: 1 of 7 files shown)


sdg/components.py
```diff
@@ -5,6 +5,26 @@
...
 
+@dsl.component(base_image=IMAGE)
+def preflight_check_op(
+    repo_branch: str,
+    repo_pr: Optional[int],
+):
+    from os import getenv
+
+    if (not repo_branch) and (repo_pr is None or repo_pr <= 0 ):
+        raise Exception("Both taxonomy repo branch and taxonomy pull request number cannot be empty")
```
MichaelClifford, line 16: This gets checked when you try to create the run. It will give an error if you try to proceed without a PR or branch. 
`Failed to create a new run: Failed to generate the ExecutionSpec: invalid pipeline job inputs: Invalid input error: input parameter repo_pr requires type double or integer, but the parameter value is not of number value type`

Can remove this check

sdg/components.py
```diff
@@ -5,6 +5,26 @@
...
+    model = getenv("model")
+    endpoint = getenv("endpoint")
+
+    if not api_key:
+        raise Exception("Model Server Auth Key is missing in kfp-model-server secret")
+    if not model:
+        raise Exception("Model name is missing in kfp-model-server configMap")
+    if not endpoint:
+        raise Exception("Model Server endpoint URL is missing in kfp-model-server configMap")
+
```
MichaelClifford, line 27: This will fail at the first step anyways if these params are not set correctly. Do we want to add in some additional checks here that may cause failures during the training and eval steps later on as well?

MichaelClifford (changes requested):

MichaelClifford:
Thanks @JudeNiroshan I made a couple of comments. Can you rebase this PR as well. Thanks!

leseb:
Please install https://docs.astral.sh/ruff/installation/ on your system and run `ruff format .` once https://github.com/redhat-et/ilab-on-ocp/pull/57 is merged. THEN rebase your PR, everything should go fine :).

PR #25, opened 2024-09-13
Split dataset processing and training

Shreyanand:
This PR decouples data processing from training run. We need them decoupled to successfully run the multi pod training where the worker nodes do not have access to the output PVC. 

This PR also adds `nodeSelector` to the Pytorch job that schedules master and worker on the same physical node which should be changed when PVC access modes are fixed.


training/run_main_ds.py
```diff
@@ -0,0 +1,182 @@
...
+    if not os.path.exists(train_args.ckpt_output_dir):
+        os.makedirs(train_args.ckpt_output_dir, exist_ok=True)
+    command = [
+        "torchrun",
+        f"--nnodes={torch_args.nnodes}",
+        f"--node_rank={torch_args.node_rank}",
+        f"--nproc_per_node={torch_args.nproc_per_node}",
+        f"--rdzv_id={torch_args.rdzv_id}",
+        f"--rdzv_endpoint={torch_args.rdzv_endpoint}",
+        f"training/src/instructlab/training/main_ds.py",
```
MichaelClifford, line 29: You probably want this to be its location in the container right? This will only work if you've cloned the repo which we are no longer doing.

I think `/usr/local/lib/python3.11/site-packages/instructlab/training/src/instructlab/training/main_ds.py` where you'd fine it.

MichaelClifford (changes requested):

training/run_data_processing.py
```diff
@@ -0,0 +1,82 @@
...
+    parser.add_argument("--max_seq_len", default=4096)
+    parser.add_argument("--max_batch_len", default=20000)
+
+
+    parser.add_argument("--num_epochs", default=2)
+    parser.add_argument("--effective_batch_size", default=3840)
+    parser.add_argument("--save_samples", default=0)
+    parser.add_argument("--learning_rate", default=2e-6)
+    parser.add_argument("--warmup_steps", default=800)
+    parser.add_argument("--is_padding_free", default=True)
```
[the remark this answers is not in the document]
Shreyanand: We don't but yeah if I remove them then it gives a validation error that arguments are missing.

training/run_main_ds.py
Shreyanand: good catch :)

training/run.py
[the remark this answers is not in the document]
Shreyanand: yes will add the updated one

training/components.py
```diff
@@ -90,12 +129,13 @@ def pytorchjob_manifest_op(
                   annotations:
                     sidecar.istio.io/inject: 'false'
                 spec:
+                  nodeSelector:
+                    kubernetes.io/hostname: wrk-6
```
MichaelClifford, line 133: Can you remove this? its very specific to our cluster

training/components.py
```diff
@@ -39,12 +76,14 @@ def pytorchjob_manifest_op(
...
```
MichaelClifford, line 80: Can you remove this?

training/components.py
```diff
@@ -4,7 +4,44 @@
 from kfp import dsl
 from typing import NamedTuple
 from utils.consts import PYTHON_IMAGE
+from typing import Optional
 
+DATA_IMAGE = 'quay.io/shanand/data_processing:0.0.2'
+
+@dsl.component(base_image=DATA_IMAGE)
+def data_processing_op(
```
MichaelClifford, line 12: Since `run_data_processing.py` is just one function. Do you think we should just embed it in this component and delete `run_data_processing.py`. This  will match how we do it in `sdg/components.py`. Also you will not need to copy anything into the data processing image or require users to have this very specific image.

training/data_processing/Containerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN pip install --no-cache-dir instructlab-training@git+https://github.com/instructlab/training.git
+
+USER 1001
+
+# Set HF_HOME to volatile location. We do not preserve cache etc over restarts with this container.
+# https://huggingface.co/docs/huggingface_hub/en/guides/manage-cache
+ENV HF_HOME="/tmp"
+
+# Make sure the containerfile has access to the run_data_processing.py file
+COPY run_data_processing.py .
```
MichaelClifford, line 12: Can possibly remove if we make the proposed change above.

training/components.py
Shreyanand: I can, but removing this would not force the master and the worker pods to be on the same node. That would give the PVC attach error.

PR #34, opened 2024-09-20
add MT_BENCH to pipeline

sallyom:
This adds MT_Bench - and uses ilab image with `ilab model evaluate --benchmark mt_bench`


eval/mt_bench/components.py
```diff
@@ -0,0 +1,99 @@
...
+    all_mt_bench_data = []
+    for model_name in models_list:
+        model_path = f"{models_path_prefix}/{model_name}"
+
+        #ilab model evaluate --benchmark mt_bench
+        #    --model $ILAB_MODELS_DIR/granite-7b-lab-Q4_K_M.gguf
+        #    --judge-model $ILAB_MODELS_DIR/granite-7b-lab-Q4_K_M.gguf
+
+        command = ["ilab", "model", "evaluate", "--benchmark", "mt_bench", "--model", "{models_path_prefix}/{model_name}",  "--judge-model", "{judge_model_path}"]
+        result = subprocess.run(command, capture_output=True, text=True)
```
MichaelClifford, line 63: If you are going to use the CLI, it might work to change this from a component to a container_component? Something like:

```
@dsl.container_component
def run_mt_bench_op(
    mt_bench_output: Output[Artifact],
    models_path_prefix: str,
    judge_model_path: str,
    models_list: List[int],
    ) -> NamedTuple("outputs", best_model=str, best_score=float) :
    
    return dsl.ContainerSpec(
        INSTRUCTLAB_IMAGE,
        ["/bin/sh", "-c"],
        [ f"ilab model evaluate --benchmark mt_bench --model {models_path_prefix}/{model_name} --judge-model {judge_model_path}"
        ],
    )
```

MichaelClifford (changes requested):

training/components.py
```diff
@@ -106,7 +106,7 @@ def pytorchjob_manifest_op(
                     sidecar.istio.io/inject: 'false'
                 spec:
                   nodeSelector:
-                    kubernetes.io/hostname: <NODE-NAME>
+                    kubernetes.io/hostname: 'wrk-6'
```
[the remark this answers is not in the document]
sallyom: yup, the title of this commit is `REMOVE b4 MERGE: set NODE-NAME & mock SDG` 😄

eval/mt_bench/components.py
sallyom: will do - i was not sure if this was the approach we were going to go with, or keep with using the libraries like what I have going on[ here](https://github.com/redhat-et/ilab-on-ocp/blob/main/eval/mt_bench_pipeline.py)  - so was mostly putting this up to see if it would run.

eval/mt_bench/components.py
```diff
@@ -0,0 +1,99 @@
...
+        scores[model_path] = overall_score
+
+    with open(mt_bench_output.path, 'w') as f:
+        json.dump(all_mt_bench_data, f, indent=4)
+
+    best_model = max(scores, key=scores.get)
+    best_score = scores[best_model]
+    return Outputs(best_model, best_score)
+
+@component(base_image=TOOLBOX_IMAGE)
```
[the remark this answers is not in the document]
sallyom: python image is best for this

eval/mt_bench/components.py
```diff
@@ -0,0 +1,99 @@
...
+            report["turn_one"] = float(lines[lines.index(line) + 1].strip())
+        elif line.startswith("### TURN TWO"):
+            report["turn_two"] = float(lines[lines.index(line) + 1].strip())
+        elif line.startswith("### ERROR RATE"):
+            report["error_rate"] = float(lines[lines.index(line) + 1].strip())
+    
+    return report
+
+
+@component(base_image=INSTRUCTLAB_IMAGE)
```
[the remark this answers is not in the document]
sallyom: the CMD can be overwritten at runtime (ENTRYPOINT cannot, but CMD can) - but it is having trouble running due to permissions

eval/mt_bench/components.py
```diff
@@ -0,0 +1,99 @@
+# type: ignore
+# pylint: disable=no-value-for-parameter,import-outside-toplevel,import-error
+from typing import List, NamedTuple
+from kfp.dsl import component, Input, Output, Artifact, Model, importer
+from utils.consts import TOOLBOX_IMAGE
+
+INSTRUCTLAB_IMAGE = "registry.redhat.io/rhelai1/instructlab-nvidia-rhel9:1.1"
+
+@component(base_image=TOOLBOX_IMAGE)
```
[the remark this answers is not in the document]
sallyom: python image will be better, yup

utils/components.py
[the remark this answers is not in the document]
sallyom: yup - I was testing both of these PRs in parallel so I cherry-picked that commit here, but it will be absorbed during rebase.

training/ilab-pytorch-job.yaml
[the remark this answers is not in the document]
sallyom: whichever the right image is for training is fine by me! which is the right image?

training/deployment.yaml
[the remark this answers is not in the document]
sallyom: i'll remove this commit `08cfd37f5b7cc912b438686ef2fb6aae00a43f8f` that contains image swapping, I wasn't sure which image was the latest best working image, so kept this in a separate commit for easy switching

eval/mt_bench/components.py
```diff
@@ -44,40 +45,34 @@ def launch_vllm_server_background(model_path: str, retries: int = 60, delay: int
...
+                    if process.is_running():
+                        print(f"Forcefully killing vLLM server process with PID: {process.info['pid']}")
+                        process.kill()  # Force kill if it's still running
+                    print(f"Successfully stopped vLLM server with PID: {process.info['pid']}")
+                except psutil.NoSuchProcess:
+                    print(f"Process with PID {process.info['pid']} no longer exists.")
+                except psutil.AccessDenied:
+                    print(f"Access denied when trying to terminate process with PID {process.info['pid']}.")
+                except Exception as e:
+                    print(f"Failed to terminate process with PID {process.info['pid']}. Error: {e}")
```
leseb, line 69: It was not working in the first place because vLLM does not handle signals properly. It expects to be run from the CLI and thus stopped via SIGINT.

If you use a process group in the instantiation above you can get ride of the current pid search logic and simply do:

```python
process_group_id = os.getpgid(process.pid)
process.send_signal(signal.SIGINT)
try:
    print("Waiting for vLLM server to shut down gracefully")
    process.wait(timeout)
except subprocess.TimeoutExpired:
    print(
        f"Sending SIGKILL to vLLM server since timeout ({timeout}s) expired"
    )
    process.kill()

# Attempt to cleanup any remaining child processes
# Make sure process_group is legit (> 1) before trying
if process_group_id and process_group_id > 1:
    try:
        os.killpg(process_group_id, signal.SIGKILL)
        print("Sent SIGKILL to vLLM process group")
    except ProcessLookupError:
        print("Nothing left to clean up with the vLLM process group")
else:
    print("vLLM process group id not found")
```

eval/mt_bench/components.py
```diff
@@ -14,16 +14,19 @@ def run_mt_bench_op(
...
+    def launch_vllm_server_background(model_path: str, retries: int = 60, delay: int = 5):
         import subprocess
         import time
         import requests
 
-        subprocess.run(["python3.11", "-m", "pip", "install", "vllm"], check=True)
-
         command = f"nohup python3.11 -m vllm.entrypoints.openai.api_server --model {model_path} &"
-        process = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
leseb, line 29: You can add `start_new_session=True,` to run this in a process group.

sallyom:
@leseb how about if we merge this, then you add the PID/Kill changes in a follow-up?

eval/mt_bench/components.py
sallyom: let's add your changes in a follow-up PR! Let me be done with this one!! 😆

eval/mt_bench/components.py
sallyom: Thanks! Please add in a follow-up, I need to be done with this PR 🤣

leseb:
> @leseb how about if we merge this, then you add the PID/Kill changes in a follow-up?

Sounds good :).

eval/mt_bench/components.py
```diff
@@ -0,0 +1,171 @@
...
+    models_list: List[str],
+    mt_bench_output: Output[Artifact],
+    merge_system_user_message: bool,
+    max_workers: str = "auto",
+    device: str = None,
+) -> NamedTuple('outputs', best_model=str, best_score=float):
+
+    import subprocess
+    subprocess.run(["python3.11", "-m", "pip", "install", "vllm"], check=True)
+
```
MichaelClifford, line 21: I think this is probably a better approach to get vllm into the component image. 
```suggestion
@component(base_image=EVAL_IMAGE, packages_to_install=["vllm"]))
def run_mt_bench_op(
    models_path_prefix: str,
    models_list: List[str],
    mt_bench_output: Output[Artifact],
    merge_system_user_message: bool,
    max_workers: str = "auto",
    device: str = None,
) -> NamedTuple('outputs', best_model=str, best_score=float):


```

leseb: Can we force to use`@git+https://github.com/opendatahub-io/vllm@2024.08.01`?

eval/mt_bench/components.py
```diff
@@ -0,0 +1,171 @@
...
+    scores = {}
+    all_mt_bench_data = []
+
+    for model_name in models_list:
+        print(f"Serving candidate model: {model_name}")
+        model_path = f"{models_path_prefix}/{model_name}"
+        
+        # Launch the vLLM server and wait until it is ready
+        launch_vllm_server_background(model_path)
```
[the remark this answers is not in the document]
sallyom: I'm not sure, I was wondering the same - this is how they do it in instructlab/instructlab though - but definitely let's explore that option. @leseb do you know?

eval/mt_bench/components.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+    # TODO: Ensure vLLM server is utilizing GPU
+    def launch_vllm_server_background(model_path: str, retries: int = 60, delay: int = 5):
+        import subprocess
+        import time
+        import requests
+
+        command = f"nohup python3.11 -m vllm.entrypoints.openai.api_server --model {model_path} &"
+        subprocess.Popen(command, shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
leseb, line 27:
```suggestion
        subprocess.Popen(args=command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```

Why `subprocess.PIPE`? Both `stdout` and `sterr` are not used anywhere.

eval/mt_bench/components.py
```diff
@@ -0,0 +1,169 @@
...
+) -> NamedTuple('outputs', best_model=str, best_score=float):
+
+
+    # TODO: Ensure vLLM server is utilizing GPU
+    def launch_vllm_server_background(model_path: str, retries: int = 60, delay: int = 5):
+        import subprocess
+        import time
+        import requests
+
+        command = f"nohup python3.11 -m vllm.entrypoints.openai.api_server --model {model_path} &"
```
leseb, line 26:
```suggestion
        command = [sys.executable, "-m", "vllm.entrypoints.openai.api_server", "--model", "{model_path}"]
```
Why do you need the shell thingy? `subprocess.Popen` spawns a child process already. Please don't do this.

eval/mt_bench/components.py
```diff
@@ -0,0 +1,169 @@
+# type: ignore
+# pylint: disable=no-value-for-parameter,import-outside-toplevel,import-error
+from typing import List, NamedTuple
+from kfp.dsl import component, Input, Output, Artifact, Model, importer
+from utils.consts import PYTHON_IMAGE
+
+EVAL_IMAGE = "quay.io/sallyom/instructlab-ocp:eval"
+
+@component(base_image=EVAL_IMAGE, packages_to_install=["vllm"]))
```
leseb, line 9: can we enforce a particular version?  `@git+https://github.com/opendatahub-io/vllm@2024.08.01`

eval/mt_bench/components.py
```diff
@@ -0,0 +1,169 @@
...
+
+    os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
+    candidate_server_url = "http://localhost:8000/v1"
+
+    gpu_available = torch.cuda.is_available()
+    gpu_name = torch.cuda.get_device_name(torch.cuda.current_device()) if gpu_available else "No GPU available"
+
+    print(f"GPU Available: {gpu_available}, {gpu_name}")
+
+    if max_workers == "auto":
```
tumido, line 92: Can you please document the magic const `auto`? It may be not obvious this comes hardcoded from ILab codebase.

eval/mt_bench_pipeline.py
```diff
@@ -121,7 +120,7 @@ def run_mt_bench(
     with open(mt_bench_output.path, 'w') as f:
         json.dump(mt_bench_data, f, indent=4)
 
-@component(base_image=TOOLBOX_IMAGE)
+@component(base_image=PYTHON_IMAGE)
```
tumido, line 123: Is the `PYTHON_IMAGE` imported in this file?

eval/mt_bench/components.py
sallyom: relic from another try, another time - I'll remove those

eval/mt_bench/components.py
sallyom: that was always my plan, but wasn't convenient while iterating on the 1000 different ways I tried to get vllm working properly - I'll add that to the Containerfile

eval/mt_bench/components.py
sallyom: I thought I did that? Oh, I did that in mmlu/components

eval/mt_bench_pipeline.py
sallyom: I believe so, it would fail otherwise, no? I think I remember adding it.... looking...

eval/mt_bench/components.py
leseb: It will run in the background, that's what `subprocess.Popen` run a process as a child of current main process. :)

eval/mt_bench/components.py
sallyom: @tumido I'll update the Containerfile in a follow-up & remove the runtime install

eval/mt_bench/components.py
sallyom: added

eval/mt_bench/components.py
sallyom: yes- will do that in a follow-up - will move the pip install to the Containerfile

eval/mt_bench/components.py
sallyom: hence the `sub` in `subprocess` 😄

eval/mt_bench/components.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+    def launch_vllm_server_background(model_path: str, gpu_count: int, retries: int = 60, delay: int = 5):
+        import subprocess
+        import sys
+        import time
+        import requests
+
+        if gpu_count > 0:
+            command = [sys.executable, "-m", "vllm.entrypoints.openai.api_server", "--model", "{model_path}", "--tensor-parallel-size", "{gpu_count}"]
```
leseb, line 30:
```suggestion
            command = [sys.executable, "-m", "vllm.entrypoints.openai.api_server", "--model", model_path, "--tensor-parallel-size", gpu_count]
```

Remove the double quote since this is a variable and don't use `{}` since we are not formating strings anymore.

leseb (changes requested):

eval/mt_bench/components.py
sallyom: Right - I see the error in the run: So this is what I have now:
```
        if gpu_count > 0:
            command = [
                sys.executable,
                "-m", "vllm.entrypoints.openai.api_server",
                "--model", model_path,
                "--tensor-parallel-size",
            ]
        else:
            command = [
                sys.executable,
                "-m", "vllm.entrypoints.openai.api_server",
                "--model", model_path,
            ]

        subprocess.Popen(args=command)
```

sallyom: testing now, hopefully this works!

sallyom: this worked, all set

MichaelClifford (approved):
LGTM

PR #37, opened 2024-09-25
allow negative pr number to bypass git fetch errors

JudeNiroshan:
### Allow pipeline executor to enter negative pr number

workaround which enable to execute the `taxonomy_base` to `empty`

 It will validate the entire taxonomy by performing a diff against an empty base.
Ref: https://github.com/instructlab/instructlab?tab=readme-ov-file#-list-and-validate-your-new-data

### DIFF
--- a/pipeline.yaml
+++ b/pipeline.yaml
@@ -507,7 +507,7 @@ deploymentSpec:
         - -c
         - "\nif ! [ -x \"$(command -v pip)\" ]; then\n    python3 -m ensurepip ||\
           \ python3 -m ensurepip --user || apt-get install python3-pip\nfi\n\nPIP_DISABLE_PIP_VERSION_CHECK=1\
-          \ python3 -m pip install --quiet --no-warn-script-location 'kfp==2.8.0'\
+          \ python3 -m pip install --quiet --no-warn-script-location 'kfp==2.9.0'\
           \ '--no-deps' 'typing-extensions>=3.7.4,<5; python_version<\"3.9\"'  &&\
           \  python3 -m pip install --quiet --no-warn-script-location 'instructlab-training@git+https://github.com/instructlab/training.git'\
           \ && \"$0\" \"$@\"\n"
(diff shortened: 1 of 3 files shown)


tumido (approved):
LGTM 🚀

sdg/components.py
```diff
@@ -19,7 +19,7 @@ def git_clone_op(
             f"git clone {repo_url} {taxonomy.path} && cd {taxonomy.path} && "
             + f'if [ ! -z "{repo_branch}" ]; then '
             + f"git fetch origin {repo_branch} && git checkout {repo_branch}; "
-            + f'elif [ ! -z "{repo_pr}" ]; then '
+            + f'elif [ ! -z "{repo_pr}" and {repo_pr} -gt 0 ]; then '
```
leseb, line 22:
```suggestion
            + f'elif [ ! -z "{repo_pr}" && {repo_pr} -gt 0 ]; then '
```
`and`? in bash? :)

leseb (changes requested):

sdg/components.py
leseb: Also, a good opportunity to switch from `! -z` with `-n` :)

sdg/components.py
```diff
@@ -43,8 +43,8 @@ def sdg_op(
     endpoint = getenv("endpoint")
     client = openai.OpenAI(base_url=endpoint, api_key=api_key)
 
-    taxonomy_base = "main" if repo_branch or repo_pr else "empty"
-
+    taxonomy_base = "main" if repo_branch or repo_pr > 0 else "empty"
+
```
leseb, line 47: trailing space

pipeline.yaml
```diff
@@ -467,7 +467,7 @@ deploymentSpec:
         - -c
         - "\nif ! [ -x \"$(command -v pip)\" ]; then\n    python3 -m ensurepip ||\
           \ python3 -m ensurepip --user || apt-get install python3-pip\nfi\n\nPIP_DISABLE_PIP_VERSION_CHECK=1\
-          \ python3 -m pip install --quiet --no-warn-script-location 'kfp==2.9.0'\
+          \ python3 -m pip install --quiet --no-warn-script-location 'kfp==2.8.0'\
```
leseb, line 470: Does this have anything to do with allowing negative number? If not, could you keep in a separate OR or at least a separate commit? Thanks!

sdg/components.py
JudeNiroshan: c16d456ae504da7a5c53fb8573a784fb1e5be6c8

sdg/components.py
JudeNiroshan: 0f3ce38d013887095bb109a7a5f3eec08c007b6f

pipeline.yaml
JudeNiroshan: 3fdfb3aacbbcb0da231c6029f943f7488ab1c476

leseb:
This PR https://github.com/redhat-et/ilab-on-ocp/pull/38 switched from 2.9 to 2.8 why are we switching again?

JudeNiroshan:
> This PR #38 switched from 2.9 to 2.8 why are we switching again?

It looks like in that PR they forgot to commit pipenv.lock. As you can see in the main branch, we are still using [kfp 2.9.0](https://github.com/redhat-et/ilab-on-ocp/blob/8a0eec425a9dc77b198eada4d9d0836896fead93/Pipfile.lock#L245), but in the latest generated yaml, it is on 2.8.0.

cooktheryan (approved):

sdg/components.py
```diff
@@ -43,7 +43,7 @@ def sdg_op(
     endpoint = getenv("endpoint")
     client = openai.OpenAI(base_url=endpoint, api_key=api_key)
 
-    taxonomy_base = "main" if repo_branch or repo_pr else "empty"
+    taxonomy_base = "main" if repo_branch or repo_pr > 0 else "empty"
```
leseb, line 46: This breaks since repo_branch is string.

PR #38, opened 2024-09-26
No cache for model list task

MichaelClifford:
This PR updates the `model_list_task` to never use a the pipeline cache as it can cause an error in `run_mmlu_task` if the  list of model names in the cache to not match the actual model directory names in the PVC.


cooktheryan:
Approved based on our conversation from today

PR #39, opened 2024-09-26
update eval Containerfile, install vllm

sallyom:
This PR is for the evaluation Containerfile, it updates the installation of vllm to happen at build time rather than runtime.

### DIFF
--- a/eval/Containerfile
+++ b/eval/Containerfile
@@ -1,18 +1,26 @@
 FROM nvcr.io/nvidia/cuda:12.4.1-devel-ubi9
 
-RUN dnf install -y python3.11 python3.11-devel git python3-pip make automake gcc gcc-c++ \
+ARG EVAL_VERSION=v0.2.1
+ARG VLLM_VERSION=2024.08.01
+
+RUN dnf install -y python3.11 python3.11-devel python3-pip \
+    git make automake gcc gcc-c++ \
+    libcudnn8 libcudnn8-devel \
     && dnf clean all \
     && rm -rf /var/cache/*dnf*
 
-# TODO update to instructlab/eval after https://github.com/instructlab/eval/pull/128 or equivalent merges
+# Required for mt_bench
+ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib:/usr/local/cuda-12.4/compat"
+
+# Fix for 9.0a in 2.4 but vLLM pulls 2.3.1 - https://github.com/pytorch/pytorch/pull/123243
+ENV TORCH_CUDA_ARCH_LIST="5.0;8.0;8.6;8.7;8.9;9.0+PTX"
+
 RUN python3.11 -m ensurepip \
-    && python3.11 -m pip install --no-cache-dir git+https://github.com/instructlab/eval@main \
+    && python3.11 -m pip install vllm@git+https://github.com/opendatahub-io/vllm@${VLLM_VERSION} \
+    && python3.11 -m pip install --no-cache-dir git+https://github.com/instructlab/eval@${EVAL_VERSION} \
     && rm -rf /usr/bin/python3 && ln -s /usr/bin/python3.11 /usr/bin/python3
 
 ENV HF_HOME=/tmp
 ENV HOME=/tmp
-
-# required for mt_bench
-ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib"
 
 ENTRYPOINT ["python3.11"]
--- a/utils/consts.py
+++ b/utils/consts.py
@@ -1,3 +1,4 @@
 PYTHON_IMAGE = "registry.access.redhat.com/ubi9/python-311:latest"
 TOOLBOX_IMAGE = "registry.access.redhat.com/ubi9/toolbox"
 OC_IMAGE = "registry.redhat.io/openshift4/ose-cli"
+EVAL_IMAGE = "quay.io/sallyom/instructlab-ocp:eval-v0.1.2"
--- a/eval/mmlu/components.py
+++ b/eval/mmlu/components.py
@@ -2,9 +2,7 @@
 # pylint: disable=no-value-for-parameter,import-outside-toplevel,import-error
 from typing import List, NamedTuple
 from kfp.dsl import component, Input, Output, Artifact, Model, importer
-from utils.consts import PYTHON_IMAGE
-
-EVAL_IMAGE = "quay.io/sallyom/instructlab-ocp:eval"
+from utils.consts import PYTHON_IMAGE, EVAL_IMAGE
 
 @component(base_image=EVAL_IMAGE)
 def run_mmlu_op(
--- a/eval/mt_bench/components.py
+++ b/eval/mt_bench/components.py
@@ -2,11 +2,9 @@
 # pylint: disable=no-value-for-parameter,import-outside-toplevel,import-error
 from typing import List, NamedTuple
 from kfp.dsl import component, Input, Output, Artifact, Model, importer
-from utils.consts import PYTHON_IMAGE
+from utils.consts import PYTHON_IMAGE, EVAL_IMAGE
 
-EVAL_IMAGE = "quay.io/sallyom/instructlab-ocp:eval"
-
-@component(base_image=EVAL_IMAGE, packages_to_install=["vllm"])
+@component(base_image=EVAL_IMAGE)
 def run_mt_bench_op(
     models_path_prefix: str,
     models_list: List[str],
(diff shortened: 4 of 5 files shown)


eval/Containerfile
```diff
@@ -4,15 +4,16 @@ RUN dnf install -y python3.11 python3.11-devel git python3-pip make automake gcc
     && dnf clean all \
     && rm -rf /var/cache/*dnf*
 
-# TODO update to instructlab/eval after https://github.com/instructlab/eval/pull/128 or equivalent merges
+ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib"
```
leseb, line 7: Can we keep the "# required for mt_bench" comment?

eval/Containerfile
```diff
@@ -4,15 +4,16 @@ RUN dnf install -y python3.11 python3.11-devel git python3-pip make automake gcc
...
+
 RUN python3.11 -m ensurepip \
-    && python3.11 -m pip install --no-cache-dir git+https://github.com/instructlab/eval@main \
+    && python3.11 -m pip install packaging wheel setuptools-scm torch psutil \
+    && python3.11 -m pip install flash_attn --no-build-isolation \
+    && python3.11 -m pip install vllm@git+https://github.com/opendatahub-io/vllm@2024.08.01 \
```
leseb, line 12: How about setting a variable for both vllm and eval so next time we update, we just change the value instead of editing the installation lines (which is error prone).

Something like:

```
ENV VLLM_VERSION=2024.08.01
EVAL_VERSION=v0.2.1
```

Then in the install line do:

```
python3.11 -m pip install --no-cache-dir git+https://github.com/instructlab/eval@${EVAL_VERSION}
```

eval/Containerfile
```diff
@@ -1,18 +1,20 @@
 FROM nvcr.io/nvidia/cuda:12.4.1-devel-ubi9
 
 RUN dnf install -y python3.11 python3.11-devel git python3-pip make automake gcc gcc-c++ \
+    && dnf install -y libcudnn8 libcudnn8-devel \
```
leseb, line 4: can't this be part of the previous `dnf` call?

eval/Containerfile
sallyom: sure

eval/Containerfile
sallyom: I'll do that, but right now the vllm install is failing - i can't get it to build

eval/Containerfile
sallyom: I'm currently trying a 100 different things, bc the build is failing - you don't need to review this atm 😃

sallyom: hence the `WIP` but if you want to try building this image locally to see what the Containerfile is missing, I'd appreciate that!

eval/Containerfile
```diff
@@ -1,18 +1,20 @@
...
     && dnf clean all \
     && rm -rf /var/cache/*dnf*
 
-# TODO update to instructlab/eval after https://github.com/instructlab/eval/pull/128 or equivalent merges
+ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib"
+
+ENV TORCH_CUDA_ARCH_LIST="5.0;8.0;8.6"
```
leseb, line 10:
```suggestion
# Fix for 9.0a in 2.4 but vLLM pulls 2.3.1 - https://github.com/pytorch/pytorch/pull/123243
ENV TORCH_CUDA_ARCH_LIST="7.0 7.5 8.0 8.6 8.7 8.9 9.0+PTX"```

File: eval/Containerfile (hunk 1 above)
Comment (sallyom) [reply eval/Containerfile:10]: added it back

File: eval/Containerfile (hunk 3 above)
Comment (sallyom) [reply eval/Containerfile]: done

File: eval/Containerfile [hunk 5]
```diff
@@ -1,18 +1,24 @@
...
+ARG VLLM_VERSION=2024.08.01
+
+RUN dnf install -y python3.11 python3.11-devel python3-pip \
+    git make automake gcc gcc-c++ \
+    libcudnn8 libcudnn8-devel \
     && dnf clean all \
     && rm -rf /var/cache/*dnf*
 
-# TODO update to instructlab/eval after https://github.com/instructlab/eval/pull/128 or equivalent merges
+ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib"
```
leseb, line 12:
```suggestion
ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib:/usr/local/cuda-12.4/compat"
```

utils/consts.py
```diff
@@ -1,3 +1,4 @@
 PYTHON_IMAGE = "registry.access.redhat.com/ubi9/python-311:latest"
 TOOLBOX_IMAGE = "registry.access.redhat.com/ubi9/toolbox"
 OC_IMAGE = "registry.redhat.io/openshift4/ose-cli"
+EVAL_IMAGE = "quay.io/rcook/vllm_ilab:vllm"
```
leseb, line 4: vllm is broken since cuda lib is not in LD_LIBRARY_PATH, I'm building a new image and will push shortly.

leseb (changes requested):

eval/Containerfile
sallyom: updated, using the ARG values fro EVAL_VERSION, VLLM_VERSION

eval/Containerfile
```diff
@@ -1,18 +1,24 @@
...
+RUN dnf install -y python3.11 python3.11-devel python3-pip \
+    git make automake gcc gcc-c++ \
+    libcudnn8 libcudnn8-devel \
     && dnf clean all \
     && rm -rf /var/cache/*dnf*
 
-# TODO update to instructlab/eval after https://github.com/instructlab/eval/pull/128 or equivalent merges
+ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/local/lib/python3.11/site-packages/nvidia/cudnn/lib:/usr/local/cuda-12.4/compat"
+
+ENV TORCH_CUDA_ARCH_LIST="5.0;8.0;8.6;8.7;8.9;9.0+PTX"
```
leseb, line 14: Please add back the comment explaining why we are doing this :)

utils/consts.py
sallyom: I think we're good, it's late for you! I'm building now, too - I needed to update the list of CUDA versions ot have `;` inbetween each version, that was causing the build to fail. This works:
`ENV TORCH_CUDA_ARCH_LIST="5.0;8.0;8.6;8.7;8.9;9.0+PTX"`

leseb: 🤔 this was not failing for me

eval/Containerfile
sallyom: done

utils/consts.py
sallyom: weird - please check this latest version I pushed - I'll build it & run through the pipeline & if green, we'll merge, sg?

leseb (approved):
LGTM!

sallyom:
@leseb @cooktheryan  I'm testing with this new image & I notice when running MMLU I'm seeing that it's not using GPU, I see this:

```
[KFP Executor 2024-09-26 21:51:16,345 INFO]: PyTorch version 2.3.1 available.
/usr/local/lib64/python3.11/site-packages/torch/cuda/__init__.py:118: UserWarning: CUDA initialization: Unexpected error from cudaGetDeviceCount(). Did you run some cuda functions before calling NumCudaDevices() that might have already set an error? Error 803: system has unsupported display driver / cuda driver combination (Triggered internally at ../c10/cuda/CUDAFunctions.cpp:108.)
  return torch._C._cuda_getDeviceCount() > 0
GPU Available: False, Using: No GPU available
Running on device: 
```
The logs would normally indicate that cuda is initialized & it's using the GPU.

cooktheryan:
@sallyom is there a way to check the pod to make sure it actually launched on a GPU node

sallyom:
> @sallyom is there a way to check the pod to make sure it actually launched on a GPU node

I've never had an issue with the eval launching on a non-GPU node so it'd be very weird if it was launched on non-GPU - but looking at other runs, the training pods were left pending with `unschedulable _something something_`- so maybe a hardware issue?  - These _should_ ensure GPU node:
```
        run_mmlu_task.set_accelerator_type('nvidia.com/gpu')
        run_mmlu_task.set_accelerator_limit(1)
```

cooktheryan:
@sallyom does http://registry.redhat.io/rhelai1/instructlab-nvidia-rhel9:1.1 have the eval steps you need? It contains VLLM. I wonder if we just make the cut over now

sallyom:
I'm going to close this one for now, we'll likely switch to the ilab/ilab image instead.

PR #60, opened 2024-10-03
Add pre-commit workflow

DanNiESh:
Add pre-commit workflow and fix file formats.

To run it locally before pushing your commit to remote repo:
- Install `pre-commit` tool: `pip install pre-commit`
- From inside this repository, run `pre-commit install`
- Then run `pre-commit run --all-files` to run all these tests


leseb (commented):
I think you need to update the pipeline.yaml, can you regenerate a new one?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,28 @@
...
+      - id: detect-private-key
+
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.6.8
+    hooks:
+      # Run the linter.
+      - id: ruff
+        args: [--fix]
+      # Run the formatter.
+      - id: ruff-format
```
leseb, line 20: IMO we don't need to run the formatter here, it needs to run locally. The CI should only run something like: `ruff format --diff .` and fail if there is a diff.

leseb (commented):
Can you add something for `isort` too? Thanks!

DanNiESh:
Update: 1. added the ruff imports sorting `ruff check --select I --fix` 
2. run ruff-format check only



> I think you need to update the pipeline.yaml, can you regenerate a new one?

How should I regenerate pipeline.yaml?

leseb (approved):
LGTM! Thanks, let get this merged :)

DanNiESh:
> @DanNiESh please rebase

Done.

MichaelClifford (approved):
LGTM!
Thanks @DanNiESh

PR #62, opened 2024-10-06
ruff format standalone.py

sallyom:
this file is formatting with my other PRs, opening to clear it from those


leseb (commented):
Unfortunately, we cannot really do this at the moment, this will get reset as soon as a new change to `standalone.tpl` is made.
I've been thinking of doing some formatting from the generation itself but ruff has not lib and I'd like to avoid calling "ruff" directly.

Issue #56, opened 2024-10-03
Remove MMLU eval step from pipeline


MichaelClifford:
InstructLab has update the phased training approach here: https://github.com/instructlab/instructlab/pull/2340

The main change is that the MMLU eval is no longer run after the first phase of training. Instead, the first phase of training is run for only 7 epochs, and the model checkpoint from the 7th epoch is automatically promoted to the next phase of training. So, the new pipeline will look like `SDG-->Phase 1 training --> Phase 2 training --> MT-Bench --> Final Eval`. 

We should make the equivalent changes in our pipeline:

1. Remove the MMLU step from our Pipeline here. 
https://github.com/redhat-et/ilab-on-ocp/blob/6cfe4a5e6259732d80530bace9db8c426c6ff4d3/pipeline.py#L185-L189 

2. Automatically promote the most recent model checkpoint as the `best_model` after phase 1 training.
https://github.com/redhat-et/ilab-on-ocp/blob/6cfe4a5e6259732d80530bace9db8c426c6ff4d3/pipeline.py#L222

MichaelClifford:
closed by #68

PR #68, opened 2024-10-08
Remove MMLU step from phased training 

MichaelClifford:
Addresses #56 

This PR primarily comments out all the MMLU eval code to remove it from the pipeline. Once we confirm none of it is needed for the final eval pieces we can remove it in a follow up PR. 

This PR also adds a new function to `pytorchjob_manifest_op` that will promote the most recent model checkpoint from the first training phase as the starting model for the second training phase.

### DIFF
--- a/training/components.py
+++ b/training/components.py
@@ -85,17 +85,31 @@ def pytorchjob_manifest_op(
     input_pvc_name: str,
     output_pvc_name: str,
     name_suffix: str,
-    path_to_model: str,
+    # path_to_model: str,
     phase_name: str,
     nproc_per_node: int = 3,
     nnodes: int = 2,
 ) -> NamedTuple("outputs", manifest=str, name=str):
     import inspect
+    import os
+
+    def list_phase1_final_model():
+        model_dir = "/output/model/hf_format"
+        models = os.listdir(model_dir)
+        newest_idx = max(
+            (os.path.getmtime(model), i) for i, model in enumerate(models)
+        )[-1]
+        newest_model = models[newest_idx]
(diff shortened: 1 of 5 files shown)


standalone.tpl
```diff
@@ -282,7 +282,7 @@ def cli():
 @click.option(
     "--eval-type",
     help="Type of evaluation to run",
-    type=click.Choice(["mmlu", "mt-bench"]),
+    type=click.Choice(["mt-bench"]),
```
leseb, line 285: We could get rid of this flag entirely IMO. If we don't please add `hidden=True` to not show it in the help.

training/components.py
```diff
@@ -85,17 +85,29 @@ def pytorchjob_manifest_op(
...
     nproc_per_node: int = 3,
     nnodes: int = 2,
 ) -> NamedTuple("outputs", manifest=str, name=str):
     import inspect
+    import os
+
+    def list_phase1_final_model():
+        model_dir = "/output/model/hf_format"
+        models = os.listdir(model_dir)
+        newest_idx = max([(int(j.split("_")[-1]), i) for i, j in enumerate(models)])[-1]
```
leseb, line 99: hum, how about we use `mtime` to sort?

```suggestion
        newest_idx = max((os.path.getmtime(model), i) for i, model in enumerate(models))[-1]
```

leseb (commented):
just nits, otherwise LGTM!

training/components.py
MichaelClifford: Much more reliable approach :) Thanks!

leseb (approved):

PR #72, opened 2024-10-08
chore: Adding sample data from upstream

franciscojavierarceo:
Resolves https://github.com/redhat-et/ilab-on-ocp/issues/71 by uploading https://github.com/instructlab/training/blob/main/sample-data/train_all_pruned_SDG.jsonl


leseb (changes requested):
Why do we need this file in the repo? Why can't we just point to the one in https://github.com/instructlab/training/blob/main/sample-data/train_all_pruned_SDG.jsonl since it's essentially the same?

astefanutti:
Right, as long as we have one sample for testing it's OK, and probably better to keep it upstream. Also LFS should better be used.

franciscojavierarceo:
Cool, I'll close then.

PR #76, opened 2024-10-08
feat: remove dependency on KFP lib + eval serving endpoint support + ci

leseb:
195862b feat: remove dependency on KFP lib
7d8d314 feat: add serving endpoint details for eval
6852064 ci: add a check to ensure the pipeline is up to date
ca03343 fix: clarify the structure of the initial tarball
2b9d01e bulk commit
2eb8e1c fix: eval, do not use external deps
8a3cef9 misc: use the correct images

commit 195862b98b0451663285eb7ddfb1f3e3854225ff
Author: Sébastien Han <seb@redhat.com>
Date:   Tue Oct 8 22:41:10 2024 +0200

    feat: remove dependency on KFP lib
    
    We now use a generic python executor to call the eval and any DSL
    component functions. They are almost identical, only the signature is
    updated to not use KFP Classes.
    
    Signed-off-by: Sébastien Han <seb@redhat.com>

commit 7d8d314b1e7095117d02a407cee30d67e11a42d6
Author: Sébastien Han <seb@redhat.com>
Date:   Wed Oct 9 09:54:26 2024 +0200

    feat: add serving endpoint details for eval
    
    During the initial MT-Bench eval, we must connect to a serving endpoint.
    The script has been updated with new required flags:
    
    * `--eval-serving-endpoint`: Serving endpoint for evaluation. e.g:
      http://serving.kubeflow.svc.cluster.local:8080/v1 - **Required**
    * `--eval-serving-model-name`: The name of the model to use for evaluation. **Required**
    * `--eval-serving-model-api-key`: The API key for the model to evaluate. `EVAL_SERVING_MODEL_API_KEY`
      environment variable can be used as well. **Required**
    
    Signed-off-by: Sébastien Han <seb@redhat.com>

commit 6852064fbe39c848e5d104ecf5d015c1ba901d5c
Author: Sébastien Han <seb@redhat.com>
Date:   Wed Oct 9 09:48:25 2024 +0200

    ci: add a check to ensure the pipeline is up to date
    
    If someone pushes code changes and the pipeline.yaml has not been
    updated accordingly, the CI will fail and ask to update it and push
    again.
    
    Also, you can now generate a pipeline with `make pipeline`.
    
    Signed-off-by: Sébastien Han <seb@redhat.com>

commit ca033433da26ea73fa41837f4b43a5fa3735075d
Author: Sébastien Han <seb@redhat.com>
Date:   Wed Oct 9 14:40:27 2024 +0200

    fix: clarify the structure of the initial tarball
[...]

PR #79, opened 2024-10-09
Remove training PVC from SDG Job

astefanutti:
This removes the training PVC from the SDG Job as it's not needed, and simplifies testing the SDG data download from S3 on cluster that does not support the provisioning of `ReadWriteMany` PVC.

### DIFF
--- a/standalone/standalone.py
+++ b/standalone/standalone.py
@@ -651,7 +651,7 @@ def get_security_context() -> kubernetes.client.V1SecurityContext:
     )
 
 
-def get_sdg_vol_mount() -> kubernetes.client.V1VolumeMount:
+def get_sdg_vol_mount() -> list[kubernetes.client.V1VolumeMount]:
     """
     Get the volume mount for the SDG job.
     """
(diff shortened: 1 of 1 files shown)


leseb (changes requested):
We need the RWX PVC on that pod to copy the model into it.

astefanutti:
> We need the RWX PVC on that pod to copy the model into it.

Isn't that the PVC referred by `MODEL_PVC_NAME`? My (limited) understanding is that it's different from the one referred by `TRAINING_PVC_NAME`.

leseb:
> > We need the RWX PVC on that pod to copy the model into it.
> 
> Isn't that the PVC referred by `MODEL_PVC_NAME`? My (limited) understanding is that it's different from the one referred by `TRAINING_PVC_NAME`.

Sorry for the confusion, these are old relics of artifacts coming from the "sdg" mode (now hidden) of the script. This commit fixes that https://github.com/redhat-et/ilab-on-ocp/commit/42bcfb44a9c410571ee8dcd2a14fdb9a7957d18d.

astefanutti:
> Sorry for the confusion, these are old relics of artifacts coming from the "sdg" mode (now hidden) of the script. This commit fixes that [42bcfb4](https://github.com/redhat-et/ilab-on-ocp/commit/42bcfb44a9c410571ee8dcd2a14fdb9a7957d18d).

Ah great, I missed it. Closing this in favor of #76 then. Sorry for the noise. Thanks.

PR #80, opened 2024-10-09
Let Kubernetes default the PVC StorageClass

astefanutti:
Rely on https://kubernetes.io/docs/concepts/storage/storage-classes/#default-storageclass instead of defaulting the StorageClass client side, which is sub-optimal.


leseb (changes requested):
Sorry, please put your changes in `standalone/standalone.tpl` and run `make standalone`. Thanks!

leseb (approved):

PR #89, opened 2024-10-11
Remove pytorch env var

sallyom:
A few things I noticed when running the standalone script

/cc @MichaelClifford @leseb

### DIFF
--- a/standalone/standalone.py
+++ b/standalone/standalone.py
@@ -1473,8 +1473,6 @@ def run_mt_bench_op(
                         f"Failed to terminate process with PID {process.info['pid']}. Error: {e}"
                     )
 
-    os.environ["PYTORCH_CUDA_ALLOC_CONF"] = "expandable_segments:True"
-
     gpu_available = torch.cuda.is_available()
     gpu_name = (
         torch.cuda.get_device_name(torch.cuda.current_device())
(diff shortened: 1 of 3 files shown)


standalone/standalone.py
```diff
@@ -311,7 +311,7 @@ def upload_s3_file():
 
 if [ "$STRATEGY" == "download" ]; then
   # List top-level directories only (no nested directories)
-  top_level_dirs=$(tar --exclude='*/*' --list --file {data_pvc_mount_path}/data.tar.gz)
+  top_level_dirs=$(tar  -tzf {data_pvc_mount_path}/data.tar.gz | grep '/$')
```
leseb, line 314: I'm actually removing this part in another PR (will push soon). This takes as long as untarring so not ideal... Can you remove this? Thanks!

leseb (commented):
Just one comment about the tar inspect

standalone/standalone.py
sallyom: removed!

leseb (changes requested):
you need to edit `standalone.tpl` then run `make standalone`. Thanks

sallyom:
I don't see anything to update in standalone.tpl, I removed the tar update commit - ready for review

leseb:
> I don't see anything to update in standalone.tpl, I removed the tar update commit - ready for review

My bad this part comes from a component. All good!

leseb (approved):

PR #91, opened 2024-10-11
add fsdp cpu offload params

MichaelClifford:


leseb (commented):
does it need to be conditional? (via flag)

MichaelClifford:
I don't think it needs to be conditional at this point. You'll want to use it most of the time.

PR #93, opened 2024-10-14
chore: add missing deps to Pipfile and fix a typo

tumido:
SSIA

I'm getting familiar myself with the `standalone/*` codebase. This is a housekeeping PR to fix up things I've noticed to be out of order.

### DIFF
--- a/standalone/standalone.tpl
+++ b/standalone/standalone.tpl
@@ -15,7 +15,7 @@ Dependencies:
     click: A package for creating command-line interfaces.
 
 TODO:
-    - Make sure ressources get cleaned up after the job is done. (configmap, secret etc) using a
+    - Make sure resources get cleaned up after the job is done. (configmap, secret etc) using a
       finalizer.
     - See if we can use KServe to deploy the model and serve it for SDG Data Generation.
       kubernetes_yaml/mixtral_serve/mixtral_serve.yaml
--- a/standalone/standalone.tpl
+++ b/standalone/standalone.tpl
@@ -25,6 +25,7 @@ import base64
 import json
 import logging
 import typing
+from ast import literal_eval
 from os import path
 from urllib.parse import urlparse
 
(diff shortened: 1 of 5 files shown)


standalone/standalone.tpl
```diff
@@ -914,13 +914,16 @@ def create_sdg_job(
     exec_data_processing_op_args = """
 {{exec_data_processing_op_args}}
 """
+    exec_git_clone_op_args = json.loads("""
```
leseb, line 917: Why `json.loads`? We just need a string here.

tumido: No we don't a string, we need to load this as JSON. It wasn't a string in the template before, so it renderes directly to the Python code as JSON data:

On current main branch - this is the template:
https://github.com/redhat-et/ilab-on-ocp/blob/b72c1ffd92351e4165d029811c6bc419881bb9b3/standalone/standalone.tpl#L923

And this is what gets rendered:
https://github.com/redhat-et/ilab-on-ocp/blob/b72c1ffd92351e4165d029811c6bc419881bb9b3/standalone/standalone.py#L1045-L1047

The reason why I'm fixing it, is that the template is not valid Python code and therefore it makes Python lang server unnecessarily complain. We can cope with or without this change though, it is definitely not necessary.

leseb: `args={{exec_git_clone_op_args}}` is just executed with `command=["/bin/sh", "-c"]`. Where is the Python part you mentioned?

tumido: So.. the `tpl` file is interpreted as Python code when developing, right? It "acts" as python source code that you hack on, correct? But it is not Python code. It's a template file. However Python Lang server locally, when developing, recognises this single line as a syntax error. This error is not affecting anything else than development convenience - it's just that Python Lang server doesn't understand that it is a template file and interprets it as Python.

**I'm not talking about exec time problem** or anything like that. It just raises alarms locally when working on the `tpl` file.

image

In simple terms - since `args={{exec_git_clone_op_args}}` is not valid Python code, the lang server complains about a syntax error in the template.

My aim here was to get rid of that error by working around it through the same templating shorthands you did before - just this time we don't actually want to end up with a string data in the Python file, but with an array.

leseb: Oh ok, I missed your first mention of the Python lang server. I also have the same issue locally but I've learned to ignore it 😅 . I still don't see why you need `json.loads()`. I would rather use the following:

```python
    from ast import literal_eval
    exec_git_clone_op_args = literal_eval("""
{{exec_git_clone_op_args}}
""")
```

With this, we should get a `list`.
What do you think?

Thanks!

leseb:
I think you need to ruff the imports.

leseb (approved):

PR #97, opened 2024-10-16
fix: re-establish watcher dropped connections

leseb:
8694036 chore: address pylint issues
7f08c26 fix: re-establish watcher dropped connections

commit 8694036fec705dac34eeeb246c8127ffb5c32c79
Date:   Wed Oct 16 14:40:41 2024 +0200

    chore: address pylint issues
    
    Pylint was complaining about line lenght.
    
commit 7f08c26f3fbe2670221ec71d26188bd9bd56193f
Date:   Wed Oct 16 14:55:24 2024 +0200

    fix: re-establish watcher dropped connections
    
    From time to time, the watcher connection will drop, this will leave the
    workflow hanging forever to catch an event that will never arrive.
    Upon exception catching of:
    
    ```
    Unexpected error: ("Connection broken: InvalidChunkLength(got length b'', 0 bytes read)", InvalidChunkLength(got length b'', 0 bytes read))
    ```
    
    We re-establish the watcher's connection.
    We also configured a timeout of 1m for each watcher.
    
    Once, the informers are implemented in the kubernetes python library it
    will be more robust to switch that. This is tracked here:
    https://github.com/kubernetes-client/python/issues/868.
    
### DIFF
--- a/standalone/standalone.tpl
+++ b/standalone/standalone.tpl
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+# pylint: disable=too-many-lines
 
 """
 Standalone Distributed training script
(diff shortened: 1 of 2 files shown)


sallyom (approved):
LGTM, but will defer to @MichaelClifford for final review with training updates - if you _do_  push anything new, there's a misspelled `librarie` in the comments referencing the k8s-client python issue.

standalone/standalone.tpl
```diff
@@ -1951,112 +2003,145 @@ def train(
...
+                                        worker_pod_success = True
+                                    if master_pod_success and worker_pod_success:
+                                        logger.info(
+                                            "All PytorchJob Pods completed successfully"
+                                        )
+                                        w.stop()
+                                        exit_flag = True
+                                        # Break here to avoid going into other conditions, we are
+                                        # done
+                                        break
```
MichaelClifford, line 2112: I won't block merging on this as we want to get this watcher issue resolved, but it's worth noting here: There will always be a single master pod, but in an actual situation there might be many more worker pods. This looks like it will trigger `w.stop()` and `exit_flag = True` once the first of N worker pods completes. But we actually want to wait for all worker pods to finish first.

MichaelClifford:
Small comment we can address in a later PR. Otherwise LGTM, thanks @leseb

standalone/standalone.tpl
leseb: Good point, will address next.

leseb: https://github.com/redhat-et/ilab-on-ocp/pull/103

PR #82, opened 2024-10-10
eval: add proper selector and permissions

leseb:
d92ecfd feat: add FSDP to standalone
ce439bf eval: add proper selector and permissions

commit d92ecfda7bbd38788176e7ab946d24fc5daf6b1e
Author: Sébastien Han <seb@redhat.com>
Date:   Thu Oct 10 16:00:49 2024 +0200

    feat: add FSDP to standalone
    
    Signed-off-by: Sébastien Han <seb@redhat.com>

commit ce439bf33a7d8f6aa114092b8206cfba88a3dd09
Author: Sébastien Han <seb@redhat.com>
Date:   Thu Oct 10 13:34:10 2024 +0200

    eval: add proper selector and permissions
    
    * Add resource request to run a GPU node
    * Write the best model/score to a file: /data/mt-bench-best.txt
    
    Signed-off-by: Sébastien Han <seb@redhat.com>

PR #101, opened 2024-10-16
Update *_task.yaml files with correct test path

MichaelClifford:
Closes #82 

This PR updates `update_test_lines_in_files()` to correctly set the correct file paths in the `*_task.yaml`.

### DIFF
--- a/eval/final/components.py
+++ b/eval/final/components.py
@@ -199,32 +199,29 @@ def run_final_eval_op(
     # model evaluation are taking place in separate environments.
     def update_test_lines_in_files(base_dir):
         import os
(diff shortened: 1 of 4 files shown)


sallyom (approved):
This is a _much_ cleaner way to update the yaml. We'll keep an eye on this issue: https://github.com/instructlab/sdg/issues/312, the real fix will go into SDG, but this works for now!  
LGTM

leseb (commented):
Changes should have gone in `standalone.tpl` :/ 

EDIT: actually no, but it looks like the file wasn't formatted 🤔

PR #103, opened 2024-10-17
fix: make sure all train workers are done

leseb:
2f5ffae fix: make sure all train workers are done

commit 2f5ffae720be9bfd897e665f81582e421cc942c0
Date:   Thu Oct 17 09:43:19 2024 +0200

    fix: make sure all train workers are done
    
    Previously, we were only counting a single train worker and assumed the
    whole training was done. Now we make sure all the worker are done before
    going to the next phase.
    
### DIFF
--- a/standalone/standalone.tpl
+++ b/standalone/standalone.tpl
@@ -2117,7 +2117,7 @@ def train(
                 )
 
                 master_pod_success = False
-                worker_pod_success = False
+                workers_pod_success = {}
                 # Always start by the last condition so that if the job is completed, we can stop
                 # watching If we don't do this, we might get 'stuck' into the Running condition and
                 # never stop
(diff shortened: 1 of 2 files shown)


standalone/standalone.tpl
```diff
@@ -2183,8 +2183,13 @@ def train(
                                             "Pod '%s' completed successfully",
                                             pod_event.metadata.name,
                                         )
-                                        worker_pod_success = True
-                                    if master_pod_success and worker_pod_success:
+                                        # Add the worker pod to the list of successful pods
+                                        workers_pod_success = {
+                                            pod_event.metadata.name: True
+                                        }
```
tumido, line 2189: This is not a list and will always be `len(...) == 1`, what's the point in doing it this way?

leseb: My bad, I meant appending... Thanks for catching this!

tumido (approved):
/lgtm
/approve

PR #104, opened 2024-10-17
bulk: house keeping

leseb:
24abae6 bulk: house keeping

commit 24abae683aeca56c3f37aa1a20179e6b6523cab8
Date:   Thu Oct 17 11:18:01 2024 +0200

    bulk: house keeping
    
    - re-arrange top-level section variables for readability
    - use env in pytorchjbob instead of export
    - do not hardcode things for executors coming from pipeline.py and use
      string formatting with values from standalone.py instead
    - some fixes for SDG  (full SDG run not data fetch)
    - catch a broader exception for watcher when connection is broken
    
### DIFF
--- a/pipeline.py
+++ b/pipeline.py
@@ -442,12 +442,12 @@ def gen_standalone():
 
     # The list of executor names to extract details from to generate the standalone script
     executors = {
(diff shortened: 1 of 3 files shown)


leseb:
I'm still fixing the watcher exception catch, so holding this a little.

MichaelClifford (approved):
LGTM

PR #107, opened 2024-10-17
update data_dir for training in standalone.py, update data_processing_op_args in pipeline.py

sallyom:
@MichaelClifford ptal
* the data-processing-op arg fix needs to go in pipeline.py, too, otherwise `make standalone` reverts the fix
* the data_dir is wrong in standalone.py (there's no equivalent in pipeline.py afaict) - so training pods were failing bc they were looking for `data.jsonl` in the wrong path.
* found a bug in mt_bench - instead of passing the full path to a single file, mt_bench needs to use the path_prefix to list the results of all epochs - it runs against each model, _then_ will find the "best model" - it then renames "samples_xxx" of the highest scored model to "candidate_model". 

See `-n sallyom-test` for the test runs for this PR


MichaelClifford (approved):
LGTM

PR #113, opened 2024-10-18
Use knowledge data for phase 1 training and skills data for phase 2 training in standalone script

### LINKED ISSUE #86: Add option to set the dataset for each training phase.

MichaelClifford:
Closes #86 

* Updates the `data_processing_op()` to output two datasets,`skills_processed_data` and `knowledge_processed_data` instead a single `processed_data` output. These new datasets are stored in `.../processed_data/knowledge` and  `.../processed_data/skills` respectively. 
* This changes requires that the SDG data used contains both a `knowledge_train_msgs*.jsonl` file as well as a `skills_train_msgs*.jsonl` file (which it should, but may cause a failure if a small PR is used to generate the SDG data). 

* Updates the `PYTORCH_TRAINING_JOB` template in `standalone.tpl` to use the `knowledge_processed_data` for phase 1 training and `skills_processed_data` for phase 2. 

_Note_: This PR does not add this feature to the KFP, I need to update the `pytorchJob_manifest_op()` to reflect this change. However, that function has fallen a bit behind the standalone version at this point so will bring it back to parity in a later PR.

### DIFF
--- a/pipeline.py
+++ b/pipeline.py
@@ -146,12 +146,26 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
             size="1Gi",
             storage_class_name=storage_class_name,
         )
-        sdg_to_pvc_task = artifact_to_pvc_op(
-            data=data_processing_task.outputs["processed_data"], pvc_path="/data"
+
+        sdg_skills_to_pvc_task = artifact_to_pvc_op(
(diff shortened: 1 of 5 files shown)


sallyom (approved):
this code looks clean, let's merge it - with the caveat that it hasn't been tested in the pipeline so there may be a follow-up PR to fix the pipeline - so long as this has been tested and proved to work with standalone script, I'm good with merging this! 
LGTM

Issue #114, opened 2024-10-18
Ensure evaluation steps skip *.jsonl files


MichaelClifford:
During development and testing of the standalone script, it is possible to run into the following error: 

```
OSError: It looks like the config file at '/data/model/output/phase_2/hf_format/candidate_model.jsonl' is not a valid JSON file.
```

I believe this is due to the fact that both MT_bench  and final eval expect `/data/model/output/phase_2/hf_format/` to only contain directories that include a model checkpoints. This is the case on a fresh run of the script with a new pvc. However, while developing or testing, you may have not cleaned up your PVC and there can be a number of `*.jsonl` files in that directory that the eval steps attempt to load as models causing the failure above.

We should fix this by either making sure these `*.jsonl` files are created in a different location, or that the eval steps skip over anything in that directory that is not a model checkpoint.

MichaelClifford:
closed by #116

PR #116, opened 2024-10-18
small clean up for mt_bench 

### LINKED ISSUE #115: Add check for existing `candidate_model`
During development and testing of the standalone script, it is possible to run into the following error:

```os.rename(
OSError: [Errno 39] Directory not empty: '/data/model/output/phase_2/hf_format/samples_160' -> '/data/model/output/phase_2/hf_format/candidate_model'```
```
I believe this is due to the fact a `/candidate_model` directory was created in a previous testing/development run.  This should not   be an issue for a fresh run of the script with a new PVC. However, while developing or testing, you may have not cleaned up your PVC causing the failure above. 

I think we can fix this by adding a check `run_mt_bench_op()` to see if `/candiate_model` exist prior to attempting to rename and skipping the rename if it does.

MichaelClifford:
Closes #115 and #114 

This PR prevents `run_mt_bench_op()` from attempting to load any `*.jsonl` files as a model. It also checks for the existence of a `candidate_model/` directory before trying to create one.  

Both of these issues can occur if the PVC is not properly cleaned between runs and can cause unexpected errors.

### DIFF
--- a/standalone/standalone.py
+++ b/standalone/standalone.py
@@ -1732,10 +1732,13 @@ def run_mt_bench_op(
 
     # Rename the best model directory to "candidate_model" for the next step
     # So we know which model to use for the final evaluation
-    os.rename(
-        os.path.join(models_path_prefix, best_model),
-        os.path.join(models_path_prefix, "candidate_model"),
-    )
+    if os.path.exists(os.path.join(models_path_prefix, "candidate_model")):
+        print("candidate_model already exists. Skipping renaming")
+    else:
+        os.rename(
+            os.path.join(models_path_prefix, best_model),
+            os.path.join(models_path_prefix, "candidate_model"),
+        )
 
     return outputs(best_model=best_model, best_score=best_score)
 """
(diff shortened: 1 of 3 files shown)


standalone/standalone.py
```diff
@@ -1678,6 +1678,8 @@ def shutdown_vllm(process: subprocess.Popen, timeout: int = 20):
             usable_cpu_count = multiprocessing.cpu_count() // 2
         max_workers = usable_cpu_count
 
+    # modify model_list to ignore any jsonl files present in the directory
+    models_list = [model for model in models_list if model.endswith(".jsonl") != True]
```
leseb, line 1682: Strange `if not model.endswith(".jsonl")]` was not propaged here, did you update manually?

leseb: https://github.com/redhat-et/ilab-on-ocp/pull/117

MichaelClifford: That is odd, I must have skipped a `pipeline.py gen-standalone`. Good catch thanks!

PR #121, opened 2024-10-22
fix: improve pytorchjob in kfp

MichaelClifford:
This PR includes a number of changes to improve the pytorchjob in `training/components.py`

* Changed `phase_name` parameter to `phase_num` to align with the pytorchjob in the standalone script.
* Made all numerical parameters in the `torchrun` call configurable through the  `pytorchjob_manifest_op()` args.
* Enforced phase 1 training to be on `knowledge_processed_data/data.json` and phase 2 training to be on `skills_processed_data/data.jsonl`.
* Moved `EXPORT` calls out of the container args and put them in the correct `env:` section of the container spec. 
* Reformatted the `torchrun` args for better readability.
* Added a couple echo statements to make sure users can easily see what model and dataset are being used in each training  pod. 
* Increased the default CPU request.

### DIFF
--- a/training/components.py
+++ b/training/components.py
@@ -108,9 +108,16 @@ def pytorchjob_manifest_op(
     output_pvc_name: str,
     name_suffix: str,
     # path_to_model: str,
-    phase_name: str,
+    phase_num: int,
     nproc_per_node: int = 3,
     nnodes: int = 2,
+    num_epochs: int = 2,
+    effective_batch_size: int = 3840,
+    learning_rate: float = 1e-4,
+    num_warmup_steps: int = 800,
+    save_samples: int = 0,
+    max_batch_len: int = 20000,
+    seed: int = 42,
 ) -> NamedTuple("outputs", manifest=str, name=str):
     import inspect
     import os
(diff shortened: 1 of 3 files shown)


training/components.py
```diff
@@ -108,9 +108,16 @@ def pytorchjob_manifest_op(
...
+    phase_num: int,
     nproc_per_node: int = 3,
     nnodes: int = 2,
+    num_epochs: int = 2,
+    effective_batch_size: int = 3840,
+    learning_rate: float = 1e-4,
+    num_warmup_steps: int = 800,
+    save_samples: int = 0,
+    max_batch_len: int = 20000,
+    seed: int = 42,
```
tumido, line 120: WDYT? Should any of these be workflow params or is it fine to hardcode them into the component definition?

tumido (approved):
Just a question inline. Otherwise:

/lgtm
/approve

Shreyanand (approved):
LGTM,  Thanks for cleaning it up, it looks much better now!  +1 on replacing export statements with env vars.

training/components.py
Shreyanand: `num_epochs` as a pipeline param would be very useful for tests with higher epochs. Having said that there are already quite some changes in this PR.

MichaelClifford: I'm working on adding them now 😄  . The one thing is that this PR is required for training to run correctly in the pipeline now since the data has been split into a Skills and Knowledge directories. So it's probably best to merge this now and add the exposed params in a follow up. Unless folks are working on other stuff today and no one is blocked.

tumido: Both is fine with me. It was an open question if it even makes any sense to expose them. 😄 A follow-up PR is good solution. I'm totally fine with that.

tumido (approved):

PR #122, opened 2024-10-23
Add training args to pipeline

MichaelClifford:
This PR allows the user to configure the training args from the Data Science Pipeline UI and retains the existing default values for each. 

The only new thing here is that that we've added `num_epochs_phase_1` and `num_epochs_phase_2` so that users can differentiate the number of epochs used in each phase (I believe the recommendation from InstructLab is 7 epochs for phase 1 and 10 epochs for phase 2). All other parameters are shared between both training phases.  

This work is a follow up from a conversation in #121


Shreyanand (approved):
/lgtm

PR #126, opened 2024-10-25
fixes for pipeline - batch_size

sallyom:
changes in this PR:
* `batch_size` set to `"auto"` <-final eval was failing with value of `8` in the pipeline
    * this follows the other parameters defaulted to "auto" wherever ilab has the option
* `BASE_MODE` -> `BASE_MODEL` <- probably was a typo?
* fixes order of pvc deletion so output-pvc doesn't get deleted before model-data is xferred from it


sallyom:
[test-run](https://rhods-dashboard-redhat-ods-applications.apps.ocp-beta-test.nerc.mghpcc.org/pipelines/sallyom-test/8d7a95e4-8fdc-4723-931b-be8908b06940/48051b3f-a402-4f8e-b50b-a74d9e0a38d1/runs/fabde9ab-5103-4604-92ef-eb63eec95dd5)

MichaelClifford (changes requested):

eval/final/components.py
```diff
@@ -26,7 +26,7 @@ def run_final_eval_op(
     device: str,
     model_dtype: str,
     few_shots: int,
-    batch_size: int,
+    batch_size: str,
```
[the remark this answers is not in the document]
sallyom: updated - I tried to use Union[int,str] but kfp dsl.component can't handle that :)

sallyom:
[test-run](https://rhods-dashboard-redhat-ods-applications.apps.ocp-beta-test.nerc.mghpcc.org/experiments/ilab/02978325-e127-4bef-9ebe-4f8f2e6c0875/runs/84dba3db-e467-4ea5-aeaf-c59bbecaa091)

pipeline.py
```diff
@@ -23,12 +22,15 @@
...
+BASE_MODEL = "ibm-granite/granite-7b-base"
 BASE_MODEL_DIR = "/data/model/"  # <- "model ID for vLLM chat/completions - corresponds to path within pvc"
+
+# eval args
 MMLU_TASKS_LIST = "mmlu_anatomy,mmlu_astronomy"
 MODEL_DTYPE = "bfloat16"
 FEW_SHOTS = 5
-BATCH_SIZE = 8
+# BATCH_SIZE can also be an int, for example "8" is converted to an int in eval/final
+BATCH_SIZE = "auto"
```
MichaelClifford, line 33: do you want to make this change ins the standalone.tmp as well? https://github.com/redhat-et/ilab-on-ocp/blob/e5a53233ce1d7f0a010fea3eb9595c9861cfa881/standalone/standalone.tpl#L100

sallyom: oh shoot, yup - should I change to "auto" or "8" wdyt?

sallyom: I updated standalone.tpl (and standalone.py) to `BATCH_SIZE = "auto"`

eval/final/components.py
```diff
@@ -216,6 +216,9 @@ def branch_eval_summary_to_json(
 
     print(f"GPU Available: {gpu_available}, Using: {gpu_name}")
 
+    if batch_size.isdigit():
+        batch_size = int(batch_size)
+
```
MichaelClifford, line 221: This will fail if the user sets the batch_size as an int instead of a str. This update will check for the correct type before continuing.   

```suggestion
    if type(batch_size) == str:
        if batch_size.isdigit():
             batch_size = int(batch_size)
        elif batch_size != "auto":
           raise ValueError("Only 'auto' is a valid string for 'batch_size'") 
```

standalone/standalone.tpl
```diff
@@ -97,7 +97,7 @@ MODEL_DTYPE = "bfloat16"
 MAX_WORKERS = "auto"
 MERGE_SYSTEM_USER_MESSAGE = False
 FEW_SHOTS = 5
-BATCH_SIZE = 8
+BATCH_SIZE = "auto"
```
MichaelClifford, line 100: You need to add " " around the variable here as well https://github.com/redhat-et/ilab-on-ocp/blob/fcc26a7c6b61ee9d0f7d767369e6345dd9cb40ac/pipeline.py#L501

sallyom: done!

eval/final/components.py
sallyom: will do in a follow-up, we'll need this for other params as well

leseb (approved):
LGTM

MichaelClifford (approved):
Thanks @sallyom LGTM!

PR #128, opened 2024-10-29
feat:  allow using custom certificates for serving

leseb:
b6f70d4 feat: allow using custom certificates for serving

commit b6f70d4fbdc9aa2ce09ed9ff0e1c4210fb7e92b0
Date:   Tue Oct 29 11:51:59 2024 +0100

    feat: allow using custom certificates for serving
    
    The model serving endpoint for the judge model does not always have
    verified certificates, sometimes they are self-signed. The communication
    will be encrypted but the certificate security chain won't be valid. Now
    we have a new `--judge-serving-model-ca-cert` flag that allows use
    to use custom certificates when interacting with the judge model serving
    endpoint.
    The secret that holds the judge model serving details can be amended
    with a new property: `JUDGE_CA_CERT: "cm-ca-cert"` to point to the
    ConfigMap that contains the custom certificates bundle.
    
### DIFF
--- a/standalone/standalone.py
+++ b/standalone/standalone.py
@@ -25,6 +25,7 @@ TODO:
 import base64
 import json
 import logging
+import os
 import time
 import typing
 from ast import literal_eval
@@ -97,6 +98,9 @@ MAX_WORKERS = "auto"
 MERGE_SYSTEM_USER_MESSAGE = False
 FEW_SHOTS = 5
 BATCH_SIZE = 8
+JUDGE_CA_CERT_ENV_VAR_NAME = "JUDGE_CA_CERT_PATH"
+JUDGE_CA_CERT_PATH = "/tmp"
+JUDGE_CA_CERT_CM_KEY = "ca-bundle.crt"
 
 # TEMPLATES
 PYTORCH_TRAINING_JOB = """
@@ -656,6 +660,20 @@ def show(
     envvar="JUDGE_SERVING_MODEL_API_KEY",
 )
 @click.option(
+    "--judge-serving-model-ca-cert",
+    type=str,
+    help=(
+        "Name of the Kubernetes ConfigMap containing the serving model CA cert."
+        "The expected key name is 'ca-bundle.crt'."
+    ),
+)
+@click.option(
+    "--judge-serving-model-ca-cert-cm-key",
+    type=str,
+    help="Name of the Key in the Kubernetes ConfigMap containing the serving model CA cert.",
+    default=JUDGE_CA_CERT_CM_KEY,
+)
+@click.option(
     "--judge-serving-model-secret",
     type=str,
     envvar="JUDGE_SERVING_MODEL_SECRET",
@@ -663,9 +681,10 @@ def show(
         "Name of the Kubernetes Secret containing the judge serving model endpoint. "
         "For evaluation only. "
         "The namespace is inferred from the namespace option. "
-        "The following keys are expected: JUDGE_API_KEY, JUDGE_ENDPOINT, JUDGE_NAME "
+        "The following keys are expected: JUDGE_API_KEY, JUDGE_ENDPOINT, JUDGE_NAME"
+        "Optional keys are: JUDGE_CA_CERT, JUDGE_CA_CERT_CM_KEY"
         " (JUDGE_SERVING_MODEL_SECRET env var)"
-        "If used, the --judge-serving-model-{api-key,endpoint,name} options will be ignored."
+        "If used, --judge-serving-model-{api-key,endpoint,name,ca-cert} will be ignored."
     ),
 )
 @click.option(
@@ -811,6 +830,8 @@ def run(
     judge_serving_model_endpoint: typing.Optional[str] = None,
     judge_serving_model_name: typing.Optional[str] = None,
     judge_serving_model_api_key: typing.Optional[str] = None,
+    judge_serving_model_ca_cert: typing.Optional[str] = None,
+    judge_serving_model_ca_cert_cm_key: typing.Optional[str] = None,
     judge_serving_model_secret: typing.Optional[str] = None,
     nproc_per_node: typing.Optional[int] = 1,
     eval_type: typing.Optional[str] = None,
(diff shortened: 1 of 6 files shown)


leseb:
@HumairAK FYI

standalone/standalone.py
```diff
@@ -97,6 +97,9 @@
 MERGE_SYSTEM_USER_MESSAGE = False
 FEW_SHOTS = 5
 BATCH_SIZE = 8
+JUDGE_CA_CERT_ENV_VAR_NAME = "JUDGE_CA_CERT_PATH"
+JUDGE_CA_CERT_PATH = "/tmp/judge_ca_cert.pem"
+JUDGE_CA_CERT_SECRET_KEY = "tls.crt"
```
szaher, line 102: the key is usually one of 
- `ca.crt` 
- `service-ca.crt`
- `ca-bundle.crt` 

let's use one of them not to confuse the user

standalone/standalone.tpl
```diff
@@ -1478,6 +1469,31 @@ def create_eval_job(
         ),
     )
 
+    if judge_serving_model_ca_cert:
+        # Define the volume that references the Secret
+        secret_volume = kubernetes.client.V1Volume(
+            name="judge-ca-cert-volume",
+            secret=kubernetes.client.V1SecretVolumeSource(
```
szaher, line 1476: are we fetching the CA bundle from secret or configmap?

szaher (commented):
we are fetching the ca-bundle from a configmap not a secret. 

a small nit issue, let's name the key for the configmap either ca.crt or ca-bundle.crt or service-ca.crt instead of tls.crt not to confuse the user.

standalone/standalone.tpl
leseb: Good point, forgot to update to use a CM :)

szaher (approved):
lgtm

leseb:
Corresponding eval PR https://github.com/instructlab/eval/pull/163

standalone/standalone.py
```diff
@@ -2344,6 +2375,31 @@ def find_node_dataset_directories(base_dir: str):
...
     )
 
+    if judge_serving_model_ca_cert:
+        # Define the volume that references the Secret
+        cm_volume = kubernetes.client.V1Volume(
+            name="judge-ca-cert-volume",
+            secret=kubernetes.client.V1ConfigMapVolumeSource(
+                name=judge_serving_model_ca_cert
+            ),
+        )
```
HumairAK, line 2385: think you intended to use `config_map` instead of `secret`,  (same with the comment)

standalone/standalone.tpl
```diff
@@ -1984,15 +2001,43 @@ def sdg_data_fetch(
...
+            if not cm.data.get(JUDGE_CA_CERT_SECRET_KEY):
+                raise ValueError(
+                    f"The provided ConfigMap {judge_serving_model_ca_cert} must contain the key:"
+                    f"'{JUDGE_CA_CERT_SECRET_KEY}'."
+                )
+        except kubernetes.client.rest.ApiException as exc:
+            if exc.status == 404:
+                raise ValueError(
+                    f"ConfigMap {judge_serving_model_ca_cert} not found in namespace {namespace}."
+                ) from exc
```
HumairAK, line 2032: should be updated to be a configmap, as per https://github.com/redhat-et/ilab-on-ocp/pull/128/files#diff-624e4842e8ca66f1dd2807228518a03ceb3a00b566f2e3ce4e073be8377bfae4R665 

some bits were switched to configmap, but `v1.read_namespaced_secret` still remain, (as well as the comments, and `JUDGE_CA_CERT_SECRET_KEY` etc.

standalone/standalone.py
```diff
@@ -655,6 +658,14 @@ def show(
     ),
     envvar="JUDGE_SERVING_MODEL_API_KEY",
 )
+@click.option(
+    "--judge-serving-model-ca-cert",
+    type=str,
+    help=(
+        "Name of the Kubernetes ConfigMap containing the serving model CA cert."
+        "The expected key name is 'ca-bundle.crt'."
```
HumairAK, line 666: this should be parameterized, note that in odh/rhoai the user will need to put `odh-ca-bundle.crt`, not something we want to hardcode here

standalone/standalone.py
```diff
@@ -97,6 +97,9 @@
 MERGE_SYSTEM_USER_MESSAGE = False
 FEW_SHOTS = 5
 BATCH_SIZE = 8
+JUDGE_CA_CERT_ENV_VAR_NAME = "JUDGE_CA_CERT_PATH"
+JUDGE_CA_CERT_PATH = "/tmp/judge_ca_cert.pem"
+JUDGE_CA_CERT_SECRET_KEY = "ca-bundle.crt"
```
HumairAK, line 102: you can actually just utilize the name of the key as the file name, by doing the following: 


```yaml
volumes:
        - name: "judge-ca-cert-volume"
          configMap:
            name: ${judge-serving-model-ca-cert}
```
and then in your volume mount: 

```yaml
        volumeMounts:
          - mountPath: /some/path/           # no file name
            name: "judge-ca-cert-volume"
```

Then the full path would be: ${JUDGE_CA_CERT_PATH}/${JUDGE_CA_CERT_SECRET_KEY}

standalone/standalone.py
```diff
@@ -2260,33 +2316,33 @@ def find_node_dataset_directories(base_dir: str):
...
+                )
+            ),
+        ],
+        resources=kubernetes.client.V1ResourceRequirements(
+            requests={"cpu": "1", "nvidia.com/gpu": nproc_per_node},
+            limits={"cpu": "1", "nvidia.com/gpu": nproc_per_node},
+        ),
+    )
+
+    if eval_type == EVAL_TYPE_MT_BENCH:
```
gmfrasca, line 2344: small code cleanliness nitpick: since we're centralizing the init container, perhaps we should also simply the `container` in the same way, since the only diff between `EVAL_TYPE_MT_BENCH` and `EVAL_TYPE_FINAL` appears to be the `arg` parameter passed in

standalone/standalone.py
leseb: *Sigh.* I was hoping to dodge more flags :)

standalone/standalone.py
leseb: Thanks, I removed `sub_path` and used `/tmp`.

szaher (approved):
lgtm

MichaelClifford (approved):
LGTM

PR #129, opened 2024-10-29
feat: use venv feature from KFP

leseb:
f730354 feat: use venv feature from KFP

commit f7303548874e4eb3900d27f817a9c17fddea5eb2
Date:   Tue Oct 29 18:06:45 2024 +0100

    feat: use venv feature from KFP
    
    Since this PR merged https://github.com/kubeflow/pipelines/pull/11326
    let's consume it in our code to start testing.
    
### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,8 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-kfp = {extras = ["kubernetes"], version = "*"}
+kfp = {git = "https://github.com/kubeflow/pipelines.git", ref = "df28e891c4374f7eac98cc6a4892b6e6c35a43f2", subdirectory = "sdk/python"}
+kfp-kubernetes = "*"
 click = "*"
 ruff = "*"
 pyyaml = "*"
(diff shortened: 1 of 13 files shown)


tumido (approved):
/lgtm
/approve

leseb:
I pushed more changes PTAL, I included an env override when using the RHELAI image, which is what the pipeline code will do soon>

cooktheryan:
@leseb I'm somewhat reluctant to include this because I think we would want to ensure everything is already included in the container image that we produce. If an organization has specific firewalls setup they wouldn't be able to reach the external mirrors. I think this is completely a valid solution if we cannot get changes to the base RHEL AI image but we should go that direction first

cooktheryan:
This impacts the standalone script as well then I thought the PR was KFP specific

leseb:
We might close this if KFP part of our image and then go with https://github.com/opendatahub-io/ilab-on-ocp/pull/173

leseb:
Will re-open later if we need this again. The use_venv was introduced in KFP 2.10 which was pulled by https://github.com/opendatahub-io/ilab-on-ocp/pull/173

PR #132, opened 2024-10-29
serving with a pvc and directions

cooktheryan:
Directions and procedures to use official images for mixtral


MichaelClifford (approved):
LGTM!

PR #120, opened 2024-10-22
feat: implement SDG in cluster for standalone mode

leseb:
e9fe873 feat: implement SDG in cluster for standalone mode

commit e9fe873eaefdff4b2d54494f922e6e25abfd69ed
Author: Sébastien Han <seb@redhat.com>
Date:   Tue Oct 22 12:24:01 2024 +0200

    feat: implement SDG in cluster for standalone mode
    
    Now, the standalone script supports running the SDG process in the
    end-to-end flow.
    
    Closes: https://github.com/redhat-et/ilab-on-ocp/issues/109
    Signed-off-by: Sébastien Han <seb@redhat.com>

PR #133, opened 2024-10-29
chore: add pipeline-type config for sdg in standalone

HumairAK:
Depends on: https://github.com/redhat-et/ilab-on-ocp/pull/120

Do not merge until #120 is merged. 
Will need to be rebased after #120 is merged. 

## Description 

This change parameterized this field: https://github.com/instructlab/sdg/blob/59ba8a5bb1fc29642c4e4d79f86e429feefda15f/src/instructlab/sdg/generate_data.py#L282

Keeping `simple` as the default behavior.

### DIFF
--- a/sdg/components.py
+++ b/sdg/components.py
@@ -45,6 +45,7 @@ def sdg_op(
     model = getenv("model")
     endpoint = getenv("endpoint")
     client = openai.OpenAI(base_url=endpoint, api_key=api_key)
+    sdg_pipeline = getenv("sdg_pipeline", "simple")
 
     taxonomy_base = "main" if repo_branch or (repo_pr and int(repo_pr) > 0) else "empty"
 
@@ -64,4 +65,5 @@ def sdg_op(
         model_name=model,
         chunk_word_count=1000,
         server_ctx_size=4096,
+        pipeline=sdg_pipeline,
     )
--- a/standalone/README.md
+++ b/standalone/README.md
@@ -425,6 +425,7 @@ evaluation
 * `--dry-run`: Print the generated Kubernetes resources without executing them. **Optional** -
   Default: false.
 * `--sdg-in-cluster`: Run the SDG process inside the cluster. **Optional** - Default: false.
+* `--sdg-pipeline-type`: The pipeline type used for SDG, value must be 'simple', 'full', or a valid path to a directory. **Optional** - Default: simple
 * `--num-instructions-to-generate`: Number of instructions to generate. For SDG only. **Optional** - Default: 30.
 * `--taxonomy-repo-branch`: The branch of the taxonomy repository to use. **Optional** - Default: none.
 * `--taxonomy-repo-pr`: The PR number of the taxonomy repository to use. **Optional** - Default: none.
@@ -617,6 +618,7 @@ stringData:
   api_key: ********
   endpoint: https://mistral-sallyom.apps.ocp-beta-test.nerc.mghpcc.org/v1
   model: mistral
+  sdg_pipeline: simple
 EOF
 ```
 
@@ -625,6 +627,7 @@ The list of all mandatory keys:
 * `api_key`: The API key for the model serving endpoint - **Required**
 * `endpoint`: Serving endpoint for evaluation - **Required**
 * `model`: The name of the model to use for SDG - **Required**
+* `sdg_pipeline`: The pipeline type used for SDG, value must be 'simple', 'full', or a valid path to a directory. - **Optional**
 
 ```bash
 ./standalone.py run \
--- a/standalone/standalone.py
+++ b/standalone/standalone.py
@@ -56,6 +56,7 @@ RHELAI_IMAGE = "registry.redhat.io/rhelai1/instructlab-nvidia-rhel9:1.2"
 SDG_OBJECT_STORE_SECRET_NAME = "sdg-object-store-credentials"
 SDG_SERVING_NAME = "sdg-serving-details"
 REPO_GRANITE_7B_IMAGE = "ibm-granite/granite-7b-base"  # used by HF downloader
+SDG_DEFAULT_PIPELINE = "simple"
 
 # SDG DATA PREPROCESSING (before doing training, data has to be converted)
 MAX_SEQ_LEN = 4096
@@ -845,6 +846,11 @@ def show(
     default=False,
     is_flag=True,
 )
+@click.option(
+    "--sdg-pipeline",
+    help="The pipeline used for SDG, value must be 'simple', 'full', or a valid path to a directory.",
+    default="simple",
+)
 @click.pass_context
 def run(
     ctx: click.Context,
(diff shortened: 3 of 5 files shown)


standalone/README.md
```diff
@@ -616,6 +618,7 @@ The list of all mandatory keys:
 * `api_key`: The API key for the model serving endpoint - **Required**
 * `endpoint`: Serving endpoint for evaluation - **Required**
 * `model`: The name of the model to use for SDG - **Required**
+* `pipeline_type`: The pipeline type used for SDG, one of [simple, agentic]. - **Required**
```
leseb, line 621: Because we use `simple` as the default, I think it's confusing to mark this field as required in the doc. I would treat this as required if we didn't have a default value. What do you think?

standalone/README.md
```diff
@@ -425,6 +425,7 @@ evaluation
 * `--dry-run`: Print the generated Kubernetes resources without executing them. **Optional** -
   Default: false.
 * `--sdg-in-cluster`: Run the SDG process inside the cluster. **Optional** - Default: false.
+* `--sdg-pipeline-type`: The pipeline type used for SDG, one of [simple, agentic]. **Optional** - Deafault: simple
```
leseb, line 428: Typo: "Deafault"

standalone/README.md
HumairAK: yeah I agree, I've added a new commit so you can see the changes, let me know if this works

leseb: LGTM. Let's just squash everything before we merge :)

standalone/standalone.py
```diff
@@ -56,6 +56,7 @@
 SDG_OBJECT_STORE_SECRET_NAME = "sdg-object-store-credentials"
 SDG_SERVING_NAME = "sdg-serving-details"
 REPO_GRANITE_7B_IMAGE = "ibm-granite/granite-7b-base"  # used by HF downloader
+DEFAULT_PIPELINE_TYPE = "simple"
```
leseb, line 59: Prefix with SDG to stay consistent?

leseb (commented):
Can we clarify that this is just for standalone and not the "pipeline"?

standalone/standalone.tpl
```diff
@@ -2016,6 +2028,7 @@ def sdg(
                 "api_key": sdg_serving_model_api_key,
                 "endpoint": sdg_serving_model_endpoint,
                 "model": sdg_serving_model_name,
+                "pipeline_type": sdg_pipeline_type,
```
leseb, line 2031: You also need to account for the 

`elif sdg_serving_model_secret:` 

We must validate that the key `pipeline_type` is present and that the value is either simple or agentic

leseb (changes requested):

standalone/standalone.tpl
HumairAK: addressed

standalone/standalone.py
```diff
@@ -2873,6 +2888,16 @@ def decode_base64(data):
                 # Validate the endpoint
                 endpoint = decode_base64(secret.data.get("endpoint"))
                 validate_url(endpoint)
+
+                # Validate sdg pipeline type constraints
+                if secret.data.get("pipeline_type") and secret.data.get(
```
leseb, line 2893: This could be a programming preference so feel free to reject it but I'd personally treat both cases separately:

1. the `pipeline_type` key does not exist
2. the key exists but the value if incorrect

With this we can raise more explicit errors.


```python
if not secret.data.get("pipeline_type"):
    raise ValueError("The provided secret must contain a 'pipeline_type' key.")

if secret.data["pipeline_type"] not in ["simple", "agentic"]:
    raise ValueError(
        f"The provided secret {sdg_serving_model_secret} must contain a valid "
        f"value for 'pipeline_type', one of [simple, agentic]."
    )
```

HumairAK: @leseb my understanding from your last comment about making this optional in the secret, means that it is a valid scenario that the this value is not present, so throwing on if not secret.data.get("pipeline_type"): would mean we make this case required, no?

leseb: Actually, you're right forget about what I said, the current approach is correct, if the value is not present we use the default anyway!

leseb (approved):
LGTM, let's merge once the CI is green

standalone/standalone.py
```diff
@@ -540,6 +541,15 @@ def cli():
     """
 
 
+def validate_pipeline_type(value):
```
HumairAK, line 544: may seem awkward placement, but it needs to appear before the @click.option for --sdg-pipeline-type, 
let me know if this should go somewhere else.

standalone/standalone.py
```diff
@@ -540,6 +541,15 @@ def cli():
...
+    if value in ["simple", "full"]:
+        return value
+    elif os.path.exists(value):
```
MichaelClifford, line 547: I'm not sure we need this check. How is this handled for other click options? Also, I'm pretty sure `os.path.exists(value):` will fail in all cases since the file you are checking here is what will be called inside of the SDG container not inside the system using the CLI. If this passed it would only be by coincidence.

sdg/components.py
```diff
@@ -45,6 +45,7 @@ def sdg_op(
     model = getenv("model")
     endpoint = getenv("endpoint")
     client = openai.OpenAI(base_url=endpoint, api_key=api_key)
+    pipeline_type = getenv("pipeline_type", "simple")
```
MichaelClifford, line 48: This seems like it won't work with the standalone CLI (but I assume I'm missing something here). This only allows us to set the `pipeline_type` as an env variable. I assume this will be set correctly when using the `sdg_serving_data_details` secret with the additional 'pipeline_type' variable. But I don't see how the click.options would ever get used here. Does  `ctx.obj["sdg_pipeline_type"]` somehow handle setting it as an env?

standalone/standalone.py
HumairAK: I don't think other click options are doing this sort of a check. But very good point about the location of the callee, I'll remove it.

sdg/components.py
HumairAK: Yes this value is being fed through env vars through the sdg serving secret, just like the endpoint/model/etc. 
https://github.com/redhat-et/ilab-on-ocp/blob/95eebca3bbdb162c3a6ce292f41666aabc3c71d6/standalone/standalone.tpl#L1109

See [here](https://github.com/redhat-et/ilab-on-ocp/pull/133/files#diff-7791f5c215c18f7e52873a5c8da37c76ecf97db053b8122bafc77d570d5ca585R630) for the readme docs

sallyom (commented):
I would like to update `pipeline_type` to `sdg_pipeline` and `--sdg-pipeline-type` to `--sdg-pipeline` as long as this works and is consistent with how variables get set in the standalone script, this is good by me.

HumairAK:
we'll split https://github.com/redhat-et/ilab-on-ocp/pull/127/files into 2 PRs, and the 2nd pr will cover the changes here but via a pipeline component arg, not env var

PR #135, opened 2024-10-30
feat: allow using custom certificates for serving (backport #128)

leseb:
The model serving endpoint for the judge model does not always have verified certificates, sometimes they are self-signed. The communication will be encrypted but the certificate security chain won't be valid. Now we have a new `--judge-serving-model-ca-cert` flag that allows use to use custom certificates when interacting with the judge model serving endpoint.
The secret that holds the judge model serving details can be amended with a new property: `JUDGE_CA_CERT: "cm-ca-cert"` to point to the ConfigMap that contains the custom certificates bundle.

(cherry picked from commit e1eca22dfa055938de3998ce987af5c9fc6c6962)


leseb:
THIS IS CHERRY-PICK INTO THE PHASE 1 BRANCH.

PR #144, opened 2024-10-31
increase kubectl_wait_for_op timeout

sallyom:
with mixtral teacher model and increased size of datasets, timeout for training needs an increase


MichaelClifford (approved):
LGTM

PR #153, opened 2024-11-04
Tune pcd sampling size

MichaelClifford:
This PR allows developers to tune how much data is used from the default data skills recipe. This is useful when testing the whole iLab pipeline and model performance is not a concern.  `SAMPLING_SIZE` should be set to 1.0 in a production setting.


sallyom (approved):
this is a temporary change - to allow for short test runs
lgtm

PR #155, opened 2024-11-05
feat: pass data between tasks via PVCs

tumido:
**This PR is a draft and needs to be tested**

This PR touches a lot of things, be warned.

It transforms our current exec logic:

```mermaid
gantt
    title Before PR
    dateFormat m
    axisFormat %M
    tickInterval 1minute
    section Pod 1
    Do work:a1 , 00, 1m
    Upload output artifacts:a2, after a1, 1m
    section Pod 2
    Download input artifacts:a3, after a2, 1m
    Do work:a4, after a3, 1m
    Upload output artifacts:a5, after a4, 1m
```

To following:

```mermaid
gantt
    title After PR
    dateFormat m
    axisFormat %M
    tickInterval 1minute
    section Pod 1
    Do work:a1 , 00, 1m
    section Pod 2
    Do work:a4, after a1, 1m
    section Pod 3
    Upload output artifacts from Pod 2:a2, after a4, 1m
    section Pod 4
    Upload output artifacts from Pod 1:a5, after a1, 1m
```

It does so by replacing removing the need to upload output artifacts created by a previous task and subsequently downloading them when starting the current task. Instead it passes the data on PVC which is simply remounted to the pod of the task which needs those outputs.

This PR makes the pipeline to use following set of PVCs:

- Model PVC - stores LLM dowloaded from HF
- SDG PVC - stores taxonomy and SDG data, including preprocessed data
- Output PVC - stores newly trained models

All current output artifacts are uploaded to S3 artifact store async to the main workflow.

Also in accordance to work around the KFP caching logic, I had to add extra `.set_caching_options(False)` calls - steps now do not depend on input artifacts, therefore KFP presumes they don't change and caches their execution - which results in data not being populated on PVCs.

image

### DIFF
--- a/pipeline.py
+++ b/pipeline.py
@@ -79,8 +79,8 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
             huggingface_importer_op,
             kubectl_apply_op,
             kubectl_wait_for_op,
-            pvc_to_artifact_op,
             pvc_to_model_op,
+            pvc_to_mt_bench_op,
         )
     else:
         from training import (
@@ -93,8 +93,8 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
             huggingface_importer_op,
             kubectl_apply_op,
             kubectl_wait_for_op,
-            pvc_to_artifact_op,
             pvc_to_model_op,
+            pvc_to_mt_bench_op,
         )
 
     # Imports for evaluation
@@ -460,39 +460,36 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
         final_eval_task.set_accelerator_type("nvidia.com/gpu")
         final_eval_task.set_accelerator_limit(1)
 
-        # Technically 'output_model_task' and 'output_data_task' can happen before evaluation,
-        # however the PVC can only be mounted once, so, setting these to _after_ so the eval proceeds.
-        output_model_task = pvc_to_artifact_op(
-            pvc_path="/output/data",
+        output_model_task = pvc_to_model_op(
+            pvc_path="/output/phase_2/model/hf_format/candidate_model",
         )
-        output_model_task.after(final_eval_task)
-        output_model_task.set_caching_options(False)
-
+        output_model_task.after(run_mt_bench_task)
         mount_pvc(
             task=output_model_task,
             pvc_name=output_pvc_task.output,
-            mount_path="/output/data",
+            mount_path="/output",
         )
 
-        output_data_task = pvc_to_model_op(
-            pvc_path="/output/model",
+        output_mt_bench_task = pvc_to_mt_bench_op(
+            pvc_path="/output/mt_bench_data.json",
         )
-        output_data_task.after(final_eval_task)
-
+        output_mt_bench_task.after(run_mt_bench_task)
         mount_pvc(
-            task=output_data_task,
+            task=output_mt_bench_task,
             pvc_name=output_pvc_task.output,
-            mount_path="/output/model",
+            mount_path="/output",
         )
 
         output_pvc_delete_task = DeletePVC(pvc_name=output_pvc_task.output)
-        output_pvc_delete_task.after(output_data_task, output_model_task)
+        output_pvc_delete_task.after(
+            output_model_task, output_mt_bench_task, final_eval_task
+        )
 
         sdg_pvc_delete_task = DeletePVC(pvc_name=sdg_input_pvc_task.output)
-        sdg_pvc_delete_task.after(output_data_task, output_model_task)
+        sdg_pvc_delete_task.after(final_eval_task)
 
         model_pvc_delete_task = DeletePVC(pvc_name=model_pvc_task.output)
-        model_pvc_delete_task.after(output_data_task, output_model_task)
+        model_pvc_delete_task.after(final_eval_task)
 
         return
 
--- a/pipeline.py
+++ b/pipeline.py
@@ -27,7 +27,6 @@ GENERATED_STANDALONE_FILE_NAME = "standalone.py"
 DEFAULT_REPO_URL = "https://github.com/instructlab/taxonomy.git"
 KFP_MODEL_SERVER_CM = "sdg/kfp-model-server.yaml"
 BASE_MODEL = "ibm-granite/granite-7b-base"
-BASE_MODEL_DIR = "/data/model/"  # <- "model ID for vLLM chat/completions - corresponds to path within pvc"
 
 # eval args
 MMLU_TASKS_LIST = "mmlu_anatomy,mmlu_astronomy"
@@ -423,7 +422,7 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
             base_branch=repo_branch,
             candidate_branch=repo_branch,
             device=device,
-            base_model_dir=BASE_MODEL_DIR,
+            base_model_dir="/model",
             max_workers=max_workers,
             merge_system_user_message=merge_system_user_message,
             model_dtype=model_dtype,
@@ -440,8 +439,8 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
         )
         mount_pvc(
             task=final_eval_task,
-            pvc_name=sdg_input_pvc_task.output,
-            mount_path="/data",
+            pvc_name=model_pvc_task.output,
+            mount_path="/model",
         )
 
         use_config_map_as_env(
--- a/pipeline.py
+++ b/pipeline.py
@@ -141,7 +141,7 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
         sdg_input_pvc_task = CreatePVC(
             pvc_name_suffix="-sdg",
             access_modes=["ReadWriteMany"],
-            size="1Gi",
+            size="10Gi",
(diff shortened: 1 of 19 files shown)


standalone/standalone.py
```diff
@@ -1101,16 +1115,20 @@ def sdg_op(
 
     print("Generating synthetic dataset for:")
     print()
-    print(read_taxonomy(taxonomy, taxonomy_base))
+    print(read_taxonomy(taxonomy_path, taxonomy_base))
+
+    # Temporary measure to limit the amount of precomputed skills data used to construct the SDG dataset.
+    # Need during development to decrease training loop times and the cost of model quality.
+    set_precomputed_skills_data_ratio(sampling_size=SAMPLING_SIZE)
```
sallyom, line 1122: This can be a follow-up - but we should add a bit about _why_ this isn't configurable - and how to set to the upstream default - and point to the GH issue 
something like:
```
This isn't configurable without setting SAMPLING_SIZE in this file above, because we don't want to expose this parameter. This is for testing & development only and will be removed in the near future. See [issue #154](https://github.com/redhat-et/ilab-on-ocp/issues/154)
```

sallyom (approved):
LGTM, see note about adding more info to SAMPLING_SIZE setting.

standalone/standalone.py
tumido: This is interesting... any idea how this change in `standalone.py` happen to be? Since I didn't change anything in the `standalone.tpl` to cause this change, did I? Can it be that the base branch has `standalone.py` out of sync?

leseb: The change must be coming from a component then.

MichaelClifford: Sorry. This was from #153 not running `make standealone` before submitting my PR

pipeline.py
```diff
@@ -121,23 +138,57 @@ def pipeline(
         seed: int = SEED,
     ):
         # SDG stage
+        sdg_input_pvc_task = CreatePVC(
+            pvc_name_suffix="-sdg",
+            access_modes=["ReadWriteMany"],
+            size="1Gi",
```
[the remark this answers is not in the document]
tumido: It works for the params we use in dev.. but I agree... we can go to 10Gi, would that be enough, you thing?

standalone/standalone.py
```diff
@@ -1081,17 +1081,31 @@ def create_sdg_container(
...
     pipeline: str,
-    taxonomy: str,
-    sdg: str,
     repo_branch: Optional[str],
     repo_pr: Optional[int],
+    taxonomy_path: str = "/data/taxonomy",
+    sdg_path: str = "/data/sdg",
 ):
-    from os import getenv
+    from os import getenv, path
```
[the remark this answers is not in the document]
tumido: This seems to come from a component I didn't change. Should I revert it or what do you suggest here?

pipeline.py
```diff
@@ -514,12 +551,12 @@ def gen_standalone():
 
     # The list of executor names to extract details from to generate the standalone script
     executors = {
-        "exec-data-processing-op": 'data_processing_op(max_seq_len={MAX_SEQ_LEN}, max_batch_len={MAX_BATCH_LEN}, sdg="{DATA_PVC_SDG_PATH}", model="{DATA_PVC_MODEL_PATH}", skills_processed_data="{PREPROCESSED_DATA_PATH_SKILLS}", knowledge_processed_data="{PREPROCESSED_DATA_PATH_KNOWLEDGE}")',
-        "exec-sdg-op": 'sdg_op(num_instructions_to_generate={num_instructions_to_generate}, pipeline="{sdg_pipeline}", repo_branch="{exec_git_clone_op_repo_branch}", repo_pr={exec_git_clone_op_repo_pr}, taxonomy="{TAXONOMY_DATA_PATH}", sdg="{DATA_PVC_SDG_PATH}")',
+        "exec-data-processing-op": 'data_processing_op(max_seq_len={MAX_SEQ_LEN}, max_batch_len={MAX_BATCH_LEN}, sdg_path="{DATA_PVC_SDG_PATH}", model_path="{DATA_PVC_MODEL_PATH}", skills_path="{PREPROCESSED_DATA_SKILLS_PATH}", knowledge_path="{PREPROCESSED_DATA_KNOWLEDGE_PATH}")',
```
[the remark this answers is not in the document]
tumido: Will do.

pipeline.py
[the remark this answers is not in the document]
leseb: Yes bumping to 10 is much more realistic.

standalone/standalone.py
tumido: Yes, this was introduced in https://github.com/redhat-et/ilab-on-ocp/pull/153. Seems like `standalone.py` was not regenerated then. Is there any change required for me to do here then?

pipeline.py
```diff
@@ -514,12 +551,12 @@ def gen_standalone():
...
+        "exec-sdg-op": 'sdg_op(num_instructions_to_generate={num_instructions_to_generate}, pipeline="{sdg_pipeline}", repo_branch="{exec_git_clone_op_repo_branch}", repo_pr={exec_git_clone_op_repo_pr}, taxonomy_path="{TAXONOMY_DATA_PATH}", sdg_path="{DATA_PVC_SDG_PATH}")',
         "exec-git-clone-op": {},
-        "exec-huggingface-importer-op": 'huggingface_importer_op(repo_name="{REPO_GRANITE_7B_IMAGE}", model="{DATA_PVC_MODEL_PATH}")',
+        "exec-huggingface-importer-op": 'huggingface_importer_op(repo_name="{REPO_GRANITE_7B_IMAGE}", model_path="{DATA_PVC_MODEL_PATH}")',
```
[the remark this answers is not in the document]
tumido: I guess you wanted to point to a different line here, correct? The `exec-run-final-eval-op` line, right?

standalone/standalone.py
```diff
@@ -1682,7 +1700,7 @@ def shutdown_vllm(process: subprocess.Popen, timeout: int = 20):
         all_mt_bench_data.append(mt_bench_data)
         scores[model_path] = overall_score
 
-    with open(mt_bench_output, "w", encoding="utf-8") as f:
+    with open(f"{output_path}/mt_bench_data.json", "w", encoding="utf-8") as f:
```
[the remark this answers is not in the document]
tumido: Yeah.. I forgot to adjust for that. I'll change that. Thanks for catching this.

pipeline.py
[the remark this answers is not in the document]
leseb: Yes sorry.

pipeline.py
```diff
@@ -731,14 +731,6 @@ def change_dsl_function_to_normal_function(rendered_code: list):
 
     import re
 
-    # Regular expression to match ".path" but not "os.path"
```
leseb, line 697: Sorry it looks like this is still present in https://github.com/redhat-et/ilab-on-ocp/pull/155/files/f24a5938e1ffb93a69f19096a031d53eb4e6d2a3..a588ed0ac06ef8478a553f74341703888f2b676b#diff-624e4842e8ca66f1dd2807228518a03ceb3a00b566f2e3ce4e073be8377bfae4R2065 so we still need this code.

tumido: Reverted. 🙂

standalone/standalone.py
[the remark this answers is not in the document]
tumido: I'll do a follow up PR then. 👍

standalone/standalone.tpl
```diff
@@ -1148,7 +1148,7 @@ def create_data_job(
 {{exec_data_processing_op_command}}
 """
     exec_data_processing_op_args = f"""
-data_processing_op(max_seq_len={MAX_SEQ_LEN}, max_batch_len={MAX_BATCH_LEN}, sdg="{DATA_PVC_SDG_PATH}", model="{DATA_PVC_MODEL_PATH}", skills_processed_data="{PREPROCESSED_DATA_SKILLS_PATH}", knowledge_processed_data="{PREPROCESSED_DATA_KNOWLEDGE_PATH}")
+data_processing_op(max_seq_len={MAX_SEQ_LEN}, max_batch_len={MAX_BATCH_LEN}, sdg_path="{DATA_PVC_SDG_PATH}", model_path="{DATA_PVC_MODEL_PATH}", skills_path="{PREPROCESSED_DATA_SKILLS_PATH}", knowledge_path="{PREPROCESSED_DATA_KNOWLEDGE_PATH}")
```
leseb, line 1151: Why is this function the only one receiving an update? I believe `run_final_eval_op` needs to change too. Did I miss something?

Did you update `standalone/standalone.py` manually?

tumido: I have no idea why is this line in `standalone.tpl` and not sourced like the rest from `pipeline.py`:
https://github.com/redhat-et/ilab-on-ocp/pull/155/files#diff-6b92e3e77956f8020482ca91889a24f3411c9a4854f2aaaac30ee504ef26ede4L517-L522

I did not update `standalone.py` manually and I don't want to. 😄

tumido: image

leseb (approved):

standalone/standalone.py
tumido: Tracked in https://github.com/redhat-et/ilab-on-ocp/issues/154

PR #163, opened 2024-11-07
chore: clean up mmlu dead code

tumido:
Signed-off-by: Tomas Coufal <tcoufal@redhat.com>


leseb (commented):
Should we update the standalone too?

tumido:
@leseb done, PTAL. 🙂 👍

standalone/standalone.py
```diff
@@ -303,13 +301,10 @@
     echo "Final data tarball path: $FINAL_DATA_TAR_PATH"
     echo "Final data tarball file: $FINAL_DATA_TAR_FILE"
     echo "Archiving data before pushing to the object store"
-    # Use '--ignore-failed-read' to ignore missing files, needed when no MMLU tasks directories are found MMLU_branch is skipped
-    # So '{mmlu_branch_scores_path}' will not exist
     tar --create \
       --gzip \
       --verbose \
-      --ignore-failed-read \
```
leseb, line 311: Why is `--ignore-failed-read` gone?

leseb (commented):
I think you need to update your local env too?

standalone/standalone.py
tumido: Because of the (also removed) comment above that line. The flag was supposedly added because there was no mmlu output, the rest of the files are essential to exist. Or am I wrong here?

MichaelClifford: @tumido I think we want to keep this in. The step `MMLU_branch` is different the than `MMLU`. 🙃 And `MMLU_branch` still happens during the final eval step. So any data it produces we want to include here.

tumido: Ahh.. I understand now.. I didn't notice at first, lol 😄 Updated/reverted.

MichaelClifford (approved):
LGTM!

PR #168, opened 2024-11-08
Update model server configmaps and secrets

### LINKED ISSUE #147: use separate secrets & configmaps and/or values for judge & teacher model servers
Currently we are using the same model & server endpoint for sdg teacher model and eval judge model. In real scenarios, there will be a separate teacher model server and judge model server. Will need to update the pipeline and standalone scripts to accommodate separate models:

**For example, standalone and kfp pipeline should expect:**

configmaps:
* judge-server
* teacher-server

secrets:
* judge-server
* teacher-server

**Currently expected:**

standalone
* only 1 secret,  `judge-server-details` secret is used (sdg isn't fully implemented in standalone yet) for all server detail

kfp pipeline
* sdg uses `kfp-model-server` secret for sdg and `judge-server` secret for eval
* sdg uses `kfp-model-server` configmap and eval uses `judge-server` configmap to find the server endpoint & model-name

MichaelClifford:
Closes #147

### DIFF
--- a/pipeline.py
+++ b/pipeline.py
@@ -15,8 +15,9 @@ from kfp.kubernetes import (
     use_secret_as_env,
 )
 
-K8S_NAME = "kfp-model-server"
-JUDGE_CONFIG_MAP = "kfp-model-server"
+TEACHER_CONFIG_MAP = "teacher-server"
+TEACHER_SECRET = "teacher-server"
+JUDGE_CONFIG_MAP = "judge-server"
 JUDGE_SECRET = "judge-server"
 MOCKED_STAGES = ["sdg", "train", "eval"]
 PIPELINE_FILE_NAME = "pipeline.yaml"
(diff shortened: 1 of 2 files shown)


pipeline.py
```diff
@@ -15,8 +15,9 @@
     use_secret_as_env,
 )
 
-K8S_NAME = "kfp-model-server"
-JUDGE_CONFIG_MAP = "kfp-model-server"
+TEACHER_CONFIG_MAP = "teacher-serving-details"
+TEACHER_SECRET = "teacher-server"
+JUDGE_CONFIG_MAP = "judge-serving-details"
 JUDGE_SECRET = "judge-server"
```
tumido, line 21: I think @sallyom wanted the names to be the same for `TEACHER_*` no matter if it's a CM or Secret. Similarly for `JUDGE_*`. Or did I get the assignment wrong?

```suggestion
TEACHER_CONFIG_MAP = "teacher-server"
TEACHER_SECRET = "teacher-server"
JUDGE_CONFIG_MAP = "judge-server"
JUDGE_SECRET = "judge-server"
```

MichaelClifford: The standalone uses a single secret called `judge-serving-details` to hold all the values for interacting with the judge model. But, if we are breaking this up here anyways, maybe we don't have to follow the same naming convention.  

https://github.com/opendatahub-io/ilab-on-ocp/blob/7ab4cb9f47052c6b72cdb43e250860047c879aaf/standalone/standalone.tpl#L94

Would it be confusing to have the Configmap and Secret to have the same name?  Should it be `teacher-server-configmap` and `teacher-server-secret` or is that a bad naming convention?

sallyom: I don't have a strong opinion, but teacher-server/judge-server is easier to remember. Now that I'm looking at it, is the only way to configure these by changing the value in this file? Same with `IMAGE_PULL_SECRET`?

cooktheryan: @sallyom I'm hoping we get to ditch IMAGE_PULL_SECRET soon but it would be interesting if those were configurable values

MichaelClifford: 👍 ok. I will make them all the same name.

sallyom: (can be a follow-up, but these can be added to the pipeline_def so they can be set in the console)

MichaelClifford:
thanks @sallyom and @tumido Change made 😄

tumido (approved):
/lgtm
/approve

PR #169, opened 2024-11-08
Use image sha256 instead of tags

szaher:
Update DS image and RHELAI Image to pull images using SHA256 instead of tags.


leseb:
@szaher do you mind sending a PR to the phase-1 branch?

PR #170, opened 2024-11-08
Use image sha256 instead of tags bkp of #169

szaher:
This is a backport of #169
Update DS image and RHELAI Image to pull images using SHA256 instead of tags.


leseb (commented):
can you use `git cherry-pick -x` instead?

szaher:
closed in favor of #171

PR #173, opened 2024-11-13
fix: do not install KFP

leseb:
15767bb fix: bump kfp version and add jinja2 dep
044217d fix: do not install KFP on RHELAI

commit 15767bb64e32215d54cccc04effdaa7bedbdb1d8
Date:   Thu Nov 14 11:30:36 2024 +0100

    fix: bump kfp version and add jinja2 dep
    
    - Pull the latest KFP tag.
    - Add jinja2 as a dep in Pipfile
    
commit 044217d2b8b5d79735b4173f47df46e3fb8741cd
Date:   Wed Nov 13 14:16:08 2024 +0100

    fix: do not install KFP on RHELAI
    
    KFP is part of our image so we don't need to trigger the code
    installation KFP.
    
### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,6 +10,7 @@ ruff = "*"
 pyyaml = "*"
 kubernetes = "*"
 pre-commit = "*"
+jinja2 = "*"
 
 [dev-packages]
 
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "c990262fb7b5cf7baa7b4726e0285a53f1955829b2e7d0f08b382d704b0fd3ff"
+            "sha256": "6e7604525112ccb4f0f790aed052e848c1f2c276081ce899d691523b9570c438"
         },
         "pipfile-spec": 6,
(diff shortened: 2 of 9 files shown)


leseb:
Looking into why the python version bumped... It makes sense anyways since 3.8 is deprecated.

leseb:
> Hmm.. KFP may be part of `RHELAI_IMAGE`, but it definitely is not present in `PYTHON_IMAGE`. 🤷
> 
> I don't think we should change every step to `RHELAI_IMAGE` though, since running 8 GB image just to list a directory in 5 seconds is nonsense...

Very good point, I totally overlooked the Python image. I also think it does not make sense to use `RHELAI_IMAGE` for everything but I believe the downstream requirement is install 0 package at runtime, so we need to find a solution to solve this.

leseb:
Ok so for now, I have decided not to touch the components using the `PYTHON_IMAGE` and only applied `install_kfp_package=False` to the `RHELAI_IMAGE` image. Also opened https://github.com/opendatahub-io/ilab-on-ocp/issues/175 to track this.

tumido (approved):
/lgtm
/approve

Issue #125, opened 2024-10-24
Rename pipeline parameters for better organization on the DSP run page


MichaelClifford:
There are a lot of parameters needed for the ilab pipeline, and they are each individually relevant to the SDG, Training or Eval steps. However, the parameter names provide no hint to the users as to what part of the pipeline they control. Furthermore, DSP automatically displays them in alphabetical order. This is simply confusing and not a good user experience. See image below for an example.

We should rename all the params with an appropriate prefix (`sdg_`, `train_`, `eval_mtbench_`, and `eval_final_`) so that they will be group logically on the "Create Run" page.      

<img width="728" alt="Screenshot 2024-10-24 at 10 48 29 AM" src="https://github.com/user-attachments/assets/731fe011-4182-4516-8a01-403060d801fa">

tumido:
I addition we can use the docstring parser of the DSL to populate description. It's not much but it can help.

https://www.kubeflow.org/docs/components/pipelines/user-guides/components/additional-functionality/#component-docstring-format

```python
@dsl.pipeline(display_name='Test pipeline')
def pipeline(
    sdg_foo: str,
    train_baz: str,
):
    """This is a test pipeline.

    Args:
        sdg_foo: This is an SDG parameter Foo that does Bar.
        train_baz: This is a Training parameter Baz that does Qux.

    """
    ...
```

Which translates to IR YAML:

```
pipelineInfo:
  description: This is a test pipeline.
  displayName: Test pipeline
  name: pipeline
root:
  dag:
    ...
  inputDefinitions:
    parameters:
      sdg_foo:
        description: This is an SDG parameter Foo that does Bar.
        parameterType: STRING
      train_baz:
        description: This is a Training parameter Baz that does Qux.
        parameterType: STRING
```

Which is surfaced as following in RHOAI:

![image](https://github.com/user-attachments/assets/62ccf0b2-9010-4ffc-8779-343330b44b70)

PR #181, opened 2024-11-15
feat: rename and document pipeline parameters

tumido:
Work in progress at the moment...

## Description
Resolves https://github.com/opendatahub-io/ilab-on-ocp/issues/125

1. Some params are renamed to align the pipeline with upstream InstructLab CLI.
2. Some params are cleaned up since they are not used at all.

Descriptions sourced mostly from:
- https://github.com/instructlab/instructlab/blob/fa5d1cc7df986b418112484c033a70b9a35a02c3/src/instructlab/configuration.py
- https://github.com/instructlab/eval/blob/f4db222e45f85add39f3b64571b27fb009ff4092/src/instructlab/eval/mt_bench.py#L177

## Known issues

RHOAI does not maintain param order defined in DSL, therefore in the UI params are ordered alphabetically. This means we'll get them in following order: `eval_*`, `mt_bench_*`, `sdg_*`, `train_*`. Not sure if this is what we want, but unless we rename them, there's nothing we can do. 

Result:

https://github.com/user-attachments/assets/527ec663-f08a-450c-8382-80beec524f6d

### DIFF
--- a/pipeline.py
+++ b/pipeline.py
@@ -31,7 +31,6 @@ BASE_MODEL = "ibm-granite/granite-7b-base"
 
 # eval args
 MMLU_TASKS_LIST = "mmlu_anatomy,mmlu_astronomy"
-MODEL_DTYPE = "bfloat16"
 FEW_SHOTS = 5
 # BATCH_SIZE can also be an int, for example "8" is converted to an int in eval/final
 BATCH_SIZE = "auto"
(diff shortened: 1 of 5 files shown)


pipeline.py
```diff
@@ -111,32 +110,65 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
...
+        train_save_samples: int = SAVE_SAMPLES,
+        train_max_batch_len: int = MAX_BATCH_LEN,
+        train_seed: int = SEED,
+        # MT Bench
+        mt_bench_max_workers: str = MAX_WORKERS,
+        # Final evaluation
+        eval_few_shots: int = FEW_SHOTS,
+        eval_batch_size: str = BATCH_SIZE,
+        # Other options
+        merge_system_user_message: bool = MERGE_SYSTEM_USER_MESSAGE,
```
MichaelClifford, line 145: This paramater is used for `mt_bench`
```suggestion
        mt_bench_merge_system_user_message: bool = MERGE_SYSTEM_USER_MESSAGE,
```

pipeline.py
```diff
@@ -115,35 +114,71 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
...
+        train_num_warmup_steps_phase_1: int = NUM_WARMUP_STEPS_PHASE_1,
+        train_num_warmup_steps_phase_2: int = NUM_WARMUP_STEPS_PHASE_2,
+        train_save_samples: int = SAVE_SAMPLES,
+        train_max_batch_len: int = MAX_BATCH_LEN,
+        train_seed: int = SEED,
+        # MT Bench
+        mt_bench_max_workers: str = MAX_WORKERS,
+        # Final evaluation
+        eval_few_shots: int = FEW_SHOTS,
+        eval_batch_size: str = BATCH_SIZE,
```
MichaelClifford, line 143: should these both be `final_eval_` just to keep naming consistency with the step name?

pipeline.py
tumido: Yes, but it's used for MT Bench Branch in final eval as well:

https://github.com/opendatahub-io/ilab-on-ocp/blob/4ed3565cc053e4a65298c92c8c6556c8d3a7de67/eval/final/components.py#L371-L385

Should I split it into 2 separate parameters?

pipeline.py
```diff
@@ -111,32 +110,65 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
...
+        # Other options
+        merge_system_user_message: bool = MERGE_SYSTEM_USER_MESSAGE,
+        storage_class_name: str = "nfs-csi",
```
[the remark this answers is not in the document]
tumido: 🤷 I would vote for `k8s_` prefix if any.. but I really don't think we need to make the names any longer...

pipeline.py
MichaelClifford: yeah. I think we should separate it. We should not make the assumption that 2 similar but different process always need the same params.

pipeline.py
tumido: I have no preference... does it convey any message to the end user? After all, they will get this context from the description field anyways...
image

pipeline.py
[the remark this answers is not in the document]
MichaelClifford: My only worry is that they will stand out as not part of any category and will just appear in alphabetical order. But if its the only one, then maybe its fine and its category is just `storage_`?

MichaelClifford (approved):

PR #183, opened 2024-11-18
fix: use RHELAI_IMAGE for data processing

MichaelClifford:
There was a mismatch between instructlab training versions between data processing and training due to the `data_ processing_op` step not using the same image as training. This PR, updates the `data_processing_op()` to use the same image.


tumido (approved):
/lgtm
/approve

PR #185, opened 2024-11-18
fix: allow phase 1 and 2 training to use different parameter values

### LINKED ISSUE #184: Allow for distinct training parameters between phase 1 and phase 2
Since each training phase uses a distinct dataset, we will want to provide users the ability to tune the parameters for each phase independently.

MichaelClifford:
Closes #184 

This PR allows pipeline users to set different values for `effective_batch_size`, `num_warmup_steps` and `learning_rate` between phase 1 and phase 2 training. Since each training phase uses a distinct dataset, we will want to provide users the ability to tune the parameters for each phase independently.


tumido (approved):
/lgtm
/approve

PR #176, opened 2024-11-14
Cleanup

leseb:
25b5e60 fix: remove non-existent member
5d81409 fix: add missing import
dd784e1 fix: add encoding to open call
c29e43e fix: remove unused imports
6e8626d fix: silence unused variables
a5971ea fix: add timeout to requests.get
2b42269 fix: remove unused "device" variable
4a755a1 fix: remove non-existent arg from run_final_eval_op
46ac860 fix: gen new pipeline

commit 25b5e6016a1b88e0d621be12b33f1edc2ef5ef9c
Author: Sébastien Han <seb@redhat.com>
Date:   Thu Nov 14 13:51:23 2024 +0100

    fix: remove non-existent member
    
    `subprocess.NoSuchProcess` does not exist. The `ProcessLookupError` case
    is handled already by the subprocess package.
    
    Also, no need to catch the general Exception.
    
    Signed-off-by: Sébastien Han <seb@redhat.com>

commit 5d814096ef96c51bb086b3b64bb32b2659776613
Author: Sébastien Han <seb@redhat.com>
Date:   Thu Nov 14 14:18:23 2024 +0100

    fix: add missing import
    
    Signed-off-by: Sébastien Han <seb@redhat.com>

commit dd784e1dac45421c3e968d776e04a2cc1f96c54d
Author: Sébastien Han <seb@redhat.com>
Date:   Thu Nov 14 14:19:07 2024 +0100

    fix: add encoding to open call
    
    Signed-off-by: Sébastien Han <seb@redhat.com>

commit c29e43ed9175fbc93b0bd2195a469f256937708e
Author: Sébastien Han <seb@redhat.com>
Date:   Thu Nov 14 14:20:17 2024 +0100

    fix: remove unused imports
    
    Signed-off-by: Sébastien Han <seb@redhat.com>

commit 6e8626d47b5c6cdbd8d78c69f2d883b3fbade15f
Author: Sébastien Han <seb@redhat.com>
Date:   Thu Nov 14 14:20:57 2024 +0100

    fix: silence unused variables
    
    Signed-off-by: Sébastien Han <seb@redhat.com>

commit a5971ead410081bd386965c15907cca6d17a8bf0
Author: Sébastien Han <seb@redhat.com>
Date:   Thu Nov 14 14:24:06 2024 +0100

    fix: add timeout to requests.get
    
    Wait 10sec for a response.
    
    Signed-off-by: Sébastien Han <seb@redhat.com>

[...]

PR #192, opened 2024-11-20
Small bugfixes from standalone/e2e test run

gmfrasca:
Fixes taxonomy_base incorrectly set to 'main' even if repo_branch or repo_num are unset

- This is because the params are evaluated as `"None"`, the string, rather than `None`, the NoneType (and 0, the int, albeit this would likely get caught by the int() cast later on)

## Description
See above

## Merge criteria:

### DIFF
--- a/pipeline.py
+++ b/pipeline.py
@@ -554,7 +554,7 @@ def gen_standalone():
     # The list of executor names to extract details from to generate the standalone script
     executors = {
         "exec-data-processing-op": 'data_processing_op(max_seq_len={MAX_SEQ_LEN}, max_batch_len={MAX_BATCH_LEN}, sdg_path="{DATA_PVC_SDG_PATH}", model_path="{DATA_PVC_MODEL_PATH}", skills_path="{PREPROCESSED_DATA_SKILLS_PATH}", knowledge_path="{PREPROCESSED_DATA_KNOWLEDGE_PATH}")',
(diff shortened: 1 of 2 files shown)


pipeline.py
```diff
@@ -556,11 +556,11 @@ def gen_standalone():
     # The list of executor names to extract details from to generate the standalone script
     executors = {
         "exec-data-processing-op": 'data_processing_op(max_seq_len={MAX_SEQ_LEN}, max_batch_len={MAX_BATCH_LEN}, sdg_path="{DATA_PVC_SDG_PATH}", model_path="{DATA_PVC_MODEL_PATH}", skills_path="{PREPROCESSED_DATA_SKILLS_PATH}", knowledge_path="{PREPROCESSED_DATA_KNOWLEDGE_PATH}")',
-        "exec-sdg-op": 'sdg_op(num_instructions_to_generate={num_instructions_to_generate}, pipeline="{sdg_pipeline}", repo_branch="{exec_git_clone_op_repo_branch}", repo_pr={exec_git_clone_op_repo_pr}, taxonomy_path="{TAXONOMY_DATA_PATH}", sdg_path="{DATA_PVC_SDG_PATH}")',
+        "exec-sdg-op": 'sdg_op(num_instructions_to_generate={num_instructions_to_generate}, pipeline="{sdg_pipeline}", repo_branch="{exec_git_clone_op_repo_branch or ''}", repo_pr={exec_git_clone_op_repo_pr or 0}, taxonomy_path="{TAXONOMY_DATA_PATH}", sdg_path="{DATA_PVC_SDG_PATH}")',
```
leseb, line 559: can you explain why this is needed?

gmfrasca: mostly for `exec_git_clone_op_repo_branch`, but this gets evaluated as the string `"None"` instead of the type `None`, so the checks later on will think this value is not false-y

gmfrasca:
looks like the final_eval parameter problem was coincidentally also fixed by #176  so removing that commit from this PR

PR #193, opened 2024-11-20
feat: use python image with pre-installed kfp package

### LINKED ISSUE #175: Avoid installing KFP on components using PYTHON_IMAGE
We should avoid installing any packages at runtime and thus use an image that has the KFP package installed.

This could mean a few things:

* stop using the PYTHON_IMAGE and use RHELAI_IMAGE: simplest method but expensive since those components are small python functions
* try to avoid those components: requires further investigation

tumido:
Resolves #175

## Description
The RHOAI provided image `quay.io/modh/odh-generic-data-science-notebook` can be used for our purposes.

I've managed to obtain following tag/version details from RHOAI team:

- `v2-*` tags use Python 3.9
- `v3-*` tags use Python 3.11
- `*-2024a-*` release is first half of 2024
- `*-2024b-*` release is second half of 2024 - aka the latest release

Therefore I've chose to use `v3-2024b-*` tag and selected the most recent build at the time of producing this PR: `v3-2024b-20241111`

Pipeline execution proof:
https://rhods-dashboard-redhat-ods-applications.apps.ocp-beta-test.nerc.mghpcc.org/experiments/ilab/f3fb689f-8f38-4822-9b6c-aa1c51d8d121/runs/14759a91-3817-48bc-afce-11f424fcecc0

image

### DIFF
--- a/pipeline.yaml
+++ b/pipeline.yaml
@@ -721,9 +721,7 @@ deploymentSpec:
         - -c
         - "\nif ! [ -x \"$(command -v pip)\" ]; then\n    python3 -m ensurepip ||\
           \ python3 -m ensurepip --user || apt-get install python3-pip\nfi\n\nPIP_DISABLE_PIP_VERSION_CHECK=1\
-          \ python3 -m pip install --quiet --no-warn-script-location 'kfp==2.10.1'\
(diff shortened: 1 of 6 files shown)


leseb (approved):
Great improvement!

cooktheryan (approved):

PR #198, opened 2024-11-21
[WIP]  DO NOT MERGE! implementing conditional pipeline logic

MichaelClifford:
Experimental PR to share feasibility of making the pipeline stages optional based on user configuration

### DIFF
--- a/pipeline.py
+++ b/pipeline.py
@@ -64,6 +64,7 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
         )
     else:
         from sdg import (
+            get_training_data,
             git_clone_op,
             sdg_op,
             sdg_to_artifact_op,
@@ -102,6 +103,8 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
 
     # Imports for evaluation
     from eval.final import run_final_eval_op
+
+    ## from eval.mmlu import run_mmlu_op, load_mmlu_results_op
     from eval.mt_bench import run_mt_bench_op
     from utils import list_models_in_directory_op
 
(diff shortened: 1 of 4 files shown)


pipeline.py
```diff
@@ -180,318 +185,397 @@ def pipeline(
...
+            output_mt_bench_task.after(run_mt_bench_task)
+            mount_pvc(
+                task=output_mt_bench_task,
+                pvc_name=output_pvc_task.output,
+                mount_path="/output",
+            )
+
+        # Pipelines
+
+        # some pre-filght image that checks all params and fails if something is wrong.
```
tumido, line 533: Maybe we can revive https://github.com/opendatahub-io/ilab-on-ocp/pull/21 efforts?

pipeline.py
```diff
@@ -180,318 +185,397 @@ def pipeline(
...
+        with dsl.If(sdg_only == True and train_only == True, name="Skip Condition"):
+            output_pvc_task, sdg_input_pvc_task, model_pvc_task = create_pvcs()
+            delete_pvcs(
+                output_pvc_task=output_pvc_task,
+                sdg_input_pvc_task=sdg_input_pvc_task,
+                model_pvc_task=model_pvc_task,
+                after=output_pvc_task,
+            )
+
+        with dsl.If(sdg_only == True and train_only == False, name="SDG Only"):
```
tumido, line 544: Can we flip the toggles? It may be more intuitive that way... like remove the `*_only` meaning from it:

| Param names | Default value |
| --- | --- |
| `run_sdg` | True |
| `run_train` | True |

So it would work like this:

| `run_sdg` value | `run_train` value | Outcome |
| --- | --- | --- |
| True | True | Full run |
| True | False | Only SDG is run |
| False | True | Oly Training is run |
| False | False | Nothing happens |


WDYT?

pipeline.py
MichaelClifford: That's a good idea.

pipeline.py
MichaelClifford: yes, that is much better 😄

PR #200, opened 2024-11-21
chore: adjust sdg pcd ratio

HumairAK:
This was an env var initially introduced for testing, and this change switches this to a configurable flag for development purposes. The default behavior, when the flag is unset, is that there is no change made to the skills data recipe.

Parameterize SDG Sampling Size

## How Has This Been Tested?
Run the standalone script, in the sdg job confirm the sdg_op() call is setting the sampling size as expected (when it is unset it should be 1.0 float, when it is set it should take on the float value passed to the parameter). 

## Merge criteria:


MichaelClifford (approved):
LGTM!

Issue #152, opened 2024-11-04
switch to the granite provided by rhel ai rather than huggingface


cooktheryan:
`ilab model download --repository docker://registry.redhat.io/rhelai1/granite-7b-starter --release latest`

Transition the above command into the pipeline rather than using huggingface

tumido:
Sadly `ilab model download` makes things only complicated:
- It requires `ilab config init` or a precreated config file - we can fake it but still - it's a nuisance
- If an OCI reference is used (which [has to be prefixed with `docker://` 🤦](https://github.com/instructlab/instructlab/blob/fa6073daf084f24346af05c021feb8f1eaa4cc44/src/instructlab/utils.py#L606)), all it does is to transform the `--repository`, `--release` and `--model-dir` args into [`skopeo copy {repository}@{release} oci:/{model-dir}` command](https://github.com/instructlab/instructlab/blob/fa6073daf084f24346af05c021feb8f1eaa4cc44/src/instructlab/model/download.py#L212-L217).

IMO it would be much better if we could use https://github.com/containers/ramalama or something instead of relying on this head-scratch logic by `ilab`...

leseb:
@tumido, for a small clarification, the token is not needed if we download an image from the instructable repository. By the way, I'm curious how other projects handle the token.

tumido:
Yeah, correct. However so far we've been using `ibm-granite/granite-7b-base` as the default base model. This model can be downloaded without auth, but not via `ilab model download` (as a workaround, setting `HF_TOKEN` to invalid, stub value, works just fine here).

cooktheryan:
@tumido @MichaelClifford @sallyom based on the conversations yesterday it seems like we can be in line with phase 2 and grab the model from S3. This will require us to remove the hugging face model pull step and integrate with a the location in object storage

leseb:
> @tumido @MichaelClifford @sallyom based on the conversations yesterday it seems like we can be in line with phase 2 and grab the model from S3. This will require us to remove the hugging face model pull step and integrate with a the location in object storage

That sounds great! An additional benefit to this is that we don't need to install `huggingface_hub` in the image anymore 🥳

PR #199, opened 2024-11-21
Use Generic DS image from MODH repo

szaher:
## Description

- [ ] Using a recent MODH image with the required dependencies to run `standalone.py` script.
- [ ] update docs to add caution before using the provide NFS docker image as it should be used for test purposes only and not for production use.

## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #202, opened 2024-11-21
feat: Source base model from S3 + helper pipeline to seed it

tumido:
Resolves: https://github.com/opendatahub-io/ilab-on-ocp/issues/152

## Description

- Change the base model consumption flow to source it from S3. The pipeline now expect a full `s3://` path which the executor already has creds for so it can use `dsl.importer()`
- Adds a helper pipeline, which runs `ilab model download` and pushes it to KFP Artifact store. This pipeline is currently specific to our cluster, due to couple of outstanding bugs in KFP.

### Helper pipeline
https://rhods-dashboard-redhat-ods-applications.apps.ocp-beta-test.nerc.mghpcc.org/pipelines/ilab/ceec2489-c172-43d8-bf49-06f4c22a0dc2/1648e706-4aa5-4601-aed9-727ed1d6e65a/runs/648f36d0-e3f0-43b8-8adb-530576beb675
image

### Main pipeline
https://rhods-dashboard-redhat-ods-applications.apps.ocp-beta-test.nerc.mghpcc.org/pipelines/ilab/ceec2489-c172-43d8-bf49-06f4c22a0dc2/9386aa5c-4db9-43df-bcc8-246f31f5f240/runs/c09cadc9-cae6-4070-bce3-507519d4e235
image

image

### DIFF
--- a/pipeline.py
+++ b/pipeline.py
@@ -13,6 +13,7 @@ from kfp.kubernetes import (
     set_image_pull_secrets,
     use_config_map_as_env,
     use_secret_as_env,
+    use_secret_as_volume,
 )
 
 TEACHER_CONFIG_MAP = "teacher-server"
@@ -21,13 +22,16 @@ JUDGE_CONFIG_MAP = "judge-server"
 JUDGE_SECRET = "judge-server"
 MOCKED_STAGES = ["sdg", "train", "eval"]
 PIPELINE_FILE_NAME = "pipeline.yaml"
+IMPORTER_PIPELINE_FILE_NAME = "importer-pipeline.yaml"
 SDG_PIPELINE = "simple"
 IMAGE_PULL_SECRET = "redhat-et-ilab-botty-pull-secret"
 STANDALONE_TEMPLATE_FILE_NAME = "standalone.tpl"
 GENERATED_STANDALONE_FILE_NAME = "standalone.py"
 DEFAULT_REPO_URL = "https://github.com/instructlab/taxonomy.git"
 KFP_MODEL_SERVER_CM = "sdg/kfp-model-server.yaml"
-BASE_MODEL = "ibm-granite/granite-7b-base"
+
+# FIXME: This value is specific to ocp-beta-test.nerc.mghpcc.org cluster, `ilab` namespace. It is quite cumbersome to copypaste and remember the path every time in dev. This default value should go away once we reach feature freeze.
+BASE_MODEL = "s3://ilab-pipeline-b1d4c2b1-ab00-4e7f-b985-697bda3df385/instructlab-base-importer/648f36d0-e3f0-43b8-8adb-530576beb675/ilab-importer-op/model/granite-7b-starter"
 
 # eval args
 MMLU_TASKS_LIST = "mmlu_anatomy,mmlu_astronomy"
@@ -51,7 +55,7 @@ MAX_BATCH_LEN = 20000
 SEED = 42
 
 
-def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
+def ilab_pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
     """Wrapper for KFP pipeline, which allows for mocking individual stages."""
 
     # Imports for SDG stage
@@ -79,7 +83,7 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
             skills_processed_data_to_artifact_op,
         )
         from utils.faked import (
-            huggingface_importer_op,
+            model_to_pvc_op,
             pvc_to_model_op,
             pvc_to_mt_bench_op,
         )
@@ -91,7 +95,7 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
             skills_processed_data_to_artifact_op,
         )
         from utils import (
-            huggingface_importer_op,
+            model_to_pvc_op,
             pvc_to_model_op,
             pvc_to_mt_bench_op,
         )
@@ -239,6 +243,9 @@ def pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
         # set_image_pull_policy(sdg_task, "Always")
 
         # Training stage
+        model_source_s3_task = dsl.importer(
+            artifact_uri=sdg_base_model, artifact_class=dsl.Model
+        )
 
         # We need to pass storage_class_name as "" to use the default StorageClass, if left empty, KFP uses "standard" StorageClass.
         # 'standard' !=  default StorageClass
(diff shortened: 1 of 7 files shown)


utils/components.py
```diff
@@ -59,12 +59,23 @@ def list_models_in_directory_op(models_folder: str) -> List[str]:
...
-    snapshot_download(repo_id=repo_name, cache_dir="/tmp", local_dir=model_path)
+@dsl.container_component
+def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.Model]):
+    return dsl.ContainerSpec(
+        RHELAI_IMAGE,
+        ["/bin/sh", "-c"],
+        [
+            "echo 'Downloading model...' && "
+            + "echo '{}' > $HOME/config.yaml && "
+            + f"ilab  --config $HOME/config.yaml model download --repository {repository} --release {release} --model-dir {base_model.path}"
```
leseb, line 79:
```suggestion
            + f"ilab --config=DEFAULT model download --repository {repository} --release {release} --model-dir {base_model.path}"
```

utils/components.py
```diff
@@ -59,12 +59,23 @@ def list_models_in_directory_op(models_folder: str) -> List[str]:
...
 
-    snapshot_download(repo_id=repo_name, cache_dir="/tmp", local_dir=model_path)
+@dsl.container_component
+def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.Model]):
+    return dsl.ContainerSpec(
+        RHELAI_IMAGE,
+        ["/bin/sh", "-c"],
+        [
+            "echo 'Downloading model...' && "
+            + "echo '{}' > $HOME/config.yaml && "
```
leseb, line 78: This shouldn't be needed.

leseb (commented):
Why did you change `standalone/standalone.py`'s mode?

utils/components.py
tumido: It actually is. `ilab model download` requires `ilab config init` but at the same time, we don't need the config. So we need to fake it.

utils/components.py
tumido: See below, we need the to force it to use our config. Other option is to generate it through `ilab config init` which is an interactive prompt...

utils/components.py
leseb: I don't understand why it is needed, I can download a model by using `--config=DEFAULT`  without having a config file on my system.
What's the error you're seeing?

leseb:
> > Why did you change standalone/standalone.py's mode?
> 
> I was fighting with #206 / #199 so in order to correct it and resolve merge conflicts here, I've deleted `standalone/standalone.py` and let it re-generate via `standalone/standalone.tpl`. As a result the mode has changed, no idea why...

Can you revert the mode to 0755 then?

tumido:
I've addressed that 0755 thingy as https://github.com/opendatahub-io/ilab-on-ocp/pull/207 so we can prevent this in future. 🙂

utils/components.py
tumido: Oh.. I didn't realize `--config=DEFAULT` means something else than without `--config` specified. I didn't even know such magic value exists. It's not documented in InstructLab as far as I know... I'll try that, thanks.

tumido:
Folks, tell me what you think about the hardcoded values? Should I remove them? I don't like them either.. but the dev exp is pretty terrible without them. 😄

leseb (commented):
Just nits

utils/components.py
```diff
@@ -24,12 +24,22 @@ def pvc_to_model_op(model: dsl.Output[dsl.Model], pvc_path: str):
...
 
-    snapshot_download(repo_id=repo_name, cache_dir="/tmp", local_dir=model_path)
+
+@dsl.container_component
+def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.Model]):
+    return dsl.ContainerSpec(
+        RHELAI_IMAGE,
+        ["/bin/sh", "-c"],
+        [
+            "echo 'Downloading model...' && "
```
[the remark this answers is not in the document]
tumido: Yup, you're right. I had additional logic around it before, so it was neat to have it here, but it's not necessary anymore. 🙂

utils/components.py
```diff
@@ -24,12 +24,21 @@ def pvc_to_model_op(model: dsl.Output[dsl.Model], pvc_path: str):
...
 
-    snapshot_download(repo_id=repo_name, cache_dir="/tmp", local_dir=model_path)
+
+@dsl.container_component
+def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.Model]):
+    return dsl.ContainerSpec(
+        RHELAI_IMAGE,
+        ["/bin/sh", "-c"],
+        [
+            f"ilab --config=DEFAULT model download --repository {repository} --release {release} --model-dir {base_model.path}"
```
leseb, line 42: Last, this can just be a single line now :)

tumido: `ruff format` will undo it to 3 lines again... 🤷

leseb (approved):
Let's gooo! Good job!

MichaelClifford (approved):
🥳 LGTM

MichaelClifford:
>Folks, tell me what you think about the hardcoded values? Should I remove them? I don't like them either.. but the dev exp is pretty terrible without them. 😄

Yeah, not ideal. But don't want to block on them either. What we could do is just remove the `BASE_MODEL`'s default to your specific image, then add docs to say "if you are pulling the model you must have a hugging-face-token and a ***-ilab-pull-secret until KFP PR's are merged"

pipeline.py
```diff
@@ -21,13 +22,16 @@
...
 SDG_PIPELINE = "simple"
 IMAGE_PULL_SECRET = "redhat-et-ilab-botty-pull-secret"
 STANDALONE_TEMPLATE_FILE_NAME = "standalone.tpl"
 GENERATED_STANDALONE_FILE_NAME = "standalone.py"
 DEFAULT_REPO_URL = "https://github.com/instructlab/taxonomy.git"
 KFP_MODEL_SERVER_CM = "sdg/kfp-model-server.yaml"
-BASE_MODEL = "ibm-granite/granite-7b-base"
+
+# FIXME: This value is specific to ocp-beta-test.nerc.mghpcc.org cluster, `ilab` namespace. It is quite cumbersome to copypaste and remember the path every time in dev. This default value should go away once we reach feature freeze.
+BASE_MODEL = "s3://ilab-pipeline-b1d4c2b1-ab00-4e7f-b985-697bda3df385/instructlab-base-importer/648f36d0-e3f0-43b8-8adb-530576beb675/ilab-importer-op/model/granite-7b-starter"
```
MichaelClifford, line 34: Sorry for the post approval comment, but after reading your note about the hardcoded values... Can we set this to something less specific to the ilab namespace?

tumido:
I'm gonna address the hardcoded values in a follow-up PR. There's more values we set which are specific to our cluster and we can't ship with those anyways.. so I'll try to make it less complicated to ship as well as develop.

PR #206, opened 2024-11-22
ci: fail if standalone.py changed and standalone.tpl did not

leseb:
1d45668 ci: fail if standalone.py changed and standalone.tpl did not

commit 1d45668af0f8e79a5c5bb106286b3b31c067b521
Date:   Fri Nov 22 10:03:33 2024 +0100

    ci: fail if standalone.py changed and standalone.tpl did not
    
    This commit adds to the pre-commit hook a check to verify that changes
    to standalone.py are accompanied by corresponding updates to
    standalone.tpl. The check will fail if standalone.py is updated without
    updating standalone.tpl ONLY if no components file(s) is/are updated.
    This ensures consistency between standalone.py and standalone.tpl, and
    prompts the user to run 'make standalone' if necessary. The check can be
    bypassed by adding the 'ignore-standalone-change' label to the pull
    request.
    
### DIFF
--- a/.github/workflows/pre_commit.yaml
+++ b/.github/workflows/pre_commit.yaml
@@ -15,6 +15,8 @@ jobs:
     steps:
       - name: Check out code
         uses: actions/checkout@v4
+        with:
+          fetch-depth: 2  # Needed for HEAD to exists
 
       - name: Setup Python
         uses: actions/setup-python@v5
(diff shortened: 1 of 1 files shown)


tumido (approved):
/lgtm
/approve

❤️ Love it!

PR #207, opened 2024-11-22
chore: ensure the standalone.py is always executable

tumido:
Addresses point brought up by @leseb here: https://github.com/opendatahub-io/ilab-on-ocp/pull/202#issuecomment-2493464060

## Description
This little change ensures that we `chmod` the `standalone.py` to 755.

### DIFF
--- a/pipeline.py
+++ b/pipeline.py
@@ -503,7 +503,7 @@ def gen_standalone():
 
     Example usage: ''' $ python pipeline.py gen-standalone '''
     """
-    from os import path
+    from os import chmod, path
 
     import yaml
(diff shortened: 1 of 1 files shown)


leseb (commented):
Instead, how about doing this in `pipeline.py` with `os.chmod(standalone_script_path, 0o755)`? After this https://github.com/opendatahub-io/ilab-on-ocp/blob/c5133e276baa76fd11dd86aee29f634933c2285f/pipeline.py#L596

tumido:
Yeah, we can do that. 👍

leseb (approved):

PR #212, opened 2024-11-26
feat(cli): run pipeline from CLI directly

tumido:
Follow up to: https://github.com/opendatahub-io/ilab-on-ocp/pull/202#issuecomment-2498046040

~@MichaelClifford, @Shreyanand  can you please help me understand the default values we set here and if **ALL** of them make sense in a generic environment. I've made a check-list in code via `FIXME:` comments. Be aware that the default values are not necessarily expected to work in the MOC cluster/`ilab` namespace context.~ (Removed from the PR, we can revisit this separatelly.)

This PR allows us to submit the pipeline run directly via CLI with no drag-and-drop-in-UI-slavery anymore...

All params are optional but allows you to control things:
```bash
$ python pipeline.py run --help
Usage: pipeline.py run [OPTIONS]

  Run the pipeline immediately against current kubernetes context (cluster and
  namespace).

  Command sets expected dev-cluster friendly default values when submitting.

Options:
  --mock [sdg|train|eval]  Mock part of the pipeline
  -e, --experiment TEXT    Set KFP experiment name.
  -r, --run TEXT           Set KFP run name.
  -p, --param TEXT         Override default parameters in KEY=VALUE format.
                           Default parameters are suitable for dev cluster -
                           the MOC cluster, `ilab` namespace.
  --help                   Show this message and exit.

```

Submitting a run then looks like this:

```bash
$ python pipeline.py run
Experiment details: https://ds-pipeline-dspa-ilab.apps.ocp-beta-test.nerc.mghpcc.org/#/experiments/details/f3fb689f-8f38-4822-9b6c-aa1c51d8d121
Run details: https://ds-pipeline-dspa-ilab.apps.ocp-beta-test.nerc.mghpcc.org/#/runs/details/2e5bb117-3a3d-44ed-bd11-49864f16df45
```

### DIFF
--- a/pipeline.py
+++ b/pipeline.py
@@ -28,13 +28,8 @@ IMAGE_PULL_SECRET = "redhat-et-ilab-botty-pull-secret"
 STANDALONE_TEMPLATE_FILE_NAME = "standalone.tpl"
 GENERATED_STANDALONE_FILE_NAME = "standalone.py"
 DEFAULT_REPO_URL = "https://github.com/instructlab/taxonomy.git"
-KFP_MODEL_SERVER_CM = "sdg/kfp-model-server.yaml"
-
-# FIXME: This value is specific to ocp-beta-test.nerc.mghpcc.org cluster, `ilab` namespace. It is quite cumbersome to copypaste and remember the path every time in dev. This default value should go away once we reach feature freeze.
-BASE_MODEL = "s3://ilab-pipeline-b1d4c2b1-ab00-4e7f-b985-697bda3df385/instructlab-base-importer/648f36d0-e3f0-43b8-8adb-530576beb675/ilab-importer-op/model/granite-7b-starter"
 
 # eval args
-MMLU_TASKS_LIST = "mmlu_anatomy,mmlu_astronomy"
 FEW_SHOTS = 5
 # BATCH_SIZE can also be an int, for example "8" is converted to an int in eval/final
 BATCH_SIZE = "auto"
@@ -113,8 +108,10 @@ def ilab_pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
         # SDG phase
         sdg_repo_url: str = "https://github.com/instructlab/taxonomy.git",
         sdg_repo_branch: Optional[str] = None,
-        sdg_repo_pr: Optional[int] = None,
-        sdg_base_model: str = BASE_MODEL,
+        sdg_repo_pr: Optional[
+            int
+        ] = None,  # FIXME: https://issues.redhat.com/browse/RHOAIRFE-467
+        sdg_base_model: str = "s3://<BUCKET>/<PATH_TO_MODEL>",
         sdg_scale_factor: int = 2,  # Renamed upstream https://github.com/instructlab/instructlab/blob/f7d40f6ed5112d59132dd832bd332fa6fbbe7010/src/instructlab/configuration.py#L279-L290
         sdg_pipeline: str = SDG_PIPELINE,
         sdg_max_batch_len: int = MAX_BATCH_LEN,
@@ -142,7 +139,7 @@ def ilab_pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
         final_eval_batch_size: str = BATCH_SIZE,
         final_eval_merge_system_user_message: bool = MERGE_SYSTEM_USER_MESSAGE,
         # Other options
-        k8s_storage_class_name: str = "nfs-csi",
+        k8s_storage_class_name: str = "standard",  # FIXME: https://github.com/kubeflow/pipelines/issues/11396, https://issues.redhat.com/browse/RHOAIRFE-470
     ):
         """InstructLab pipeline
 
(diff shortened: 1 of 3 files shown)


utils/kfp_client.py
```diff
@@ -0,0 +1,55 @@
...
+from kubernetes.client import CustomObjectsApi
+from kubernetes.client.configuration import Configuration
+from kubernetes.client.exceptions import ApiException
+from kubernetes.config import list_kube_config_contexts
+from kubernetes.config.kube_config import load_kube_config
+
+
+def get_kfp_client():
+    config = Configuration()
+    load_kube_config(client_configuration=config)
```
leseb, line 14: I believe `load_kube_config()` could fail here, should we wrap this in a try/expect block? Should we fall back to in_cluster config?

pipeline.py
```diff
@@ -23,36 +24,33 @@
...
+NUM_EPOCHS_PHASE_2 = 2  # FIXME: Is this a good default value?
+EFFECTIVE_BATCH_SIZE_PHASE_1 = 3840  # FIXME: Is this a good default value?
+EFFECTIVE_BATCH_SIZE_PHASE_2 = 3840  # FIXME: Is this a good default value?
+LEARNING_RATE_PHASE_1 = 1e-4  # FIXME: Is this a good default value?
+LEARNING_RATE_PHASE_2 = 1e-4  # FIXME: Is this a good default value?
+NUM_WARMUP_STEPS_PHASE_1 = 100  # FIXME: Is this a good default value?
+NUM_WARMUP_STEPS_PHASE_2 = 100  # FIXME: Is this a good default value?
+SAVE_SAMPLES = 0  # FIXME: Is this a good default value?
+MAX_BATCH_LEN = 20000  # FIXME: Is this a good default value?
+SEED = 42  # FIXME: Should we default to this? What is a good default value? 0?
```
leseb, line 53: I don’t believe we need all the FIXMEs here. The defaults are sourced from instructlab/instructlab and are identical, so they can be trusted in my opinion.

pipeline.py
```diff
@@ -535,6 +531,54 @@ def generate_pipeline(mock):
...
+    """
+    from utils.kfp_client import get_kfp_client
+
+    client = get_kfp_client()
+
+    dev_arguments = {
+        "k8s_storage_class_name": "nfs-csi",
+        "sdg_base_model": "s3://ilab-pipeline-b1d4c2b1-ab00-4e7f-b985-697bda3df385/instructlab-base-importer/648f36d0-e3f0-43b8-8adb-530576beb675/ilab-importer-op/model/granite-7b-starter",
+        "train_nproc_per_node": 2,
+    }
```
leseb, line 564: Why are these hardcoded if we now have `--param`?

pipeline.py
leseb: Also, I feel like this is unrelated to the main change of running the pipeline from the CLI. Should we treat this separately?

leseb (commented):
How is that different from using the `kfp` [CLI](https://kubeflow-pipelines.readthedocs.io/en/sdk-2.7.0/source/cli.html)?

tumido:
@leseb 
> How is that different from using the kfp CLI?

CLI will become usable against OCP once https://github.com/kubeflow/pipelines/pull/11400 get's into `kfp` release. Before that there's no way to authenticate KFP CLI properly. And since we're locked on `2.9` at the moment, we can't benefit from it any time soon.

pipeline.py
tumido: This is here to simplify our dev workflow. You can always override them with `--param`.

pipeline.py
tumido: Maybe we should do this separately. I don't know which some from instructlab/instructlab and which don't, that's why I'm asking in the PR description about this.

utils/kfp_client.py
leseb: Ok for not to falling back to in_cluster. How about the try/expect on the `load_kube_config()` now?

pipeline.py
leseb: I suggest breaking this into two PRs. Completing the pipeline run can be prioritized and done more quickly, while the review and adjustments for all the knobs may take more time.

Just to clarify, I checked the values, and they are sourced from InstructLab. I believe they are fine as they are.

utils/kfp_client.py
tumido: Done. 👍

pipeline.py
tumido: Removed from the PR.

utils/kfp_client.py
```diff
@@ -0,0 +1,56 @@
...
+            "Installed version of Kubeflow Pipelines does not meet minimal version criteria. Use KFPv2 please."
+        ) from e
+
+    try:
+        host = dspa["status"]["components"]["apiServer"]["externalUrl"]
+    except KeyError as e:
+        raise ApiException(
+            409,
+            "DataSciencePipelines resource is not ready. Check for .status.components.apiServer",
+        ) from e
```
leseb, line 53: Does that mean we should wait and retry?

leseb (approved):

utils/kfp_client.py
tumido: I think it's good enough to notify and fail hare. If DSPA is not ready, there's usually something else going on. And this command is not supposed to be a controller or anything. I don't want to reconcile or troubleshoot DSPA... In the nutshell it means the KFP endpoint is unreachable. That's it. KFP CLI would do the same, just fail and let the user handle it. I think that's a fair assumption. 🙂

leseb: Ok thanks for the clarification!

PR #217, opened 2024-11-29
Update docs to download models from RH registry

szaher:
## Description

Update RHELAI docs to enable model downloading from OCI registries

### How to use it

To download ilab models from OCI registries following the following instructions
> [!NOTE]
> [skopeo](https://github.com/containers/skopeo/) must be installed and policy.json file present in `$HOME/.config/containers/policy.json` or `/etc/containers/policy.json`
> If you don't have skopeo policy, you can download default policy from [here](https://github.com/containers/skopeo/blob/main/default-policy.json)
  ```bash
  $ ilab model download --repository docker://registry.redhat.io/rhelai1/granite-7b-starter --release 1.2-1728663941
  $ cp -r <path-to-model-downloaded-dir>/granite-7b-starter/* s3-data/model
  ```

## Merge criteria:

### DIFF
--- a/standalone/instructlab-rhoai.md
+++ b/standalone/instructlab-rhoai.md
@@ -43,9 +43,10 @@ Before running the training and evaluation steps we must complete the following:
   ```
 
 * Download ilab model repository in s3-data model direct
+
   ```
-  $ ilab model download --repository ibm-granite/granite-7b-base
-  $ cp -r <path-to-model-downloaded-dir>/ibm-granite/granite-7b-base/* s3-data/model
+  $ ilab model download --repository docker://registry.redhat.io/rhelai1/granite-7b-starter --release 1.2-1728663941
+  $ cp -r <path-to-model-downloaded-dir>/granite-7b-starter/* s3-data/model
   ```
 
 * Clone taxonomy repository (this could be any taxonomy repo and any branches in the given repo)


standalone/instructlab-rhoai.md
```diff
@@ -43,9 +43,12 @@ Before running the training and evaluation steps we must complete the following:
   ```
 
 * Download ilab model repository in s3-data model direct
+> [!NOTE]
+> [skopeo](https://github.com/containers/skopeo/) must be installed and policy.json file present in `$HOME/.config/containers/policy.json` or `/etc/containers/policy.json`
+> If you don't have skopeo policy, you can download default policy from [here](https://github.com/containers/skopeo/blob/main/default-policy.json)
```
leseb, line 48: Do you think this should point to a RH doc instead?

szaher: do you mean something like https://github.com/containers/skopeo/blob/main/install.md ? I am not sure if we have red hat specific docs for installing skopeo?

leseb: How about https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/building_running_and_managing_containers/assembly_running-skopeo-buildah-and-podman-in-a-container#proc_running-skopeo-in-a-container_assembly_running-skopeo-buildah-and-podman-in-a-container?

szaher: @leseb I think `ilab model download` can't call skopeo from within a container. it calls it using subprocess https://github.com/instructlab/instructlab/blob/0a8dee3857f9e5e2ce9cee8691033ad9059eef34/src/instructlab/model/download.py#L212-L224 

the only available guide is https://github.com/containers/skopeo/blob/main/install.md which you can find it in the repository.

leseb: Right, I missed it was running inside a container. I guess my point is that since this is a product doc we shouldn't point to an upstream installation guide.

leseb: @HumairAK this is correct,I thought it was phase 2 and related to the earlier concern we had. I agree that since this i all running from a RHELAI box then skopeo is already available. No need to mention this IMO.

szaher: done!

PR #224, opened 2024-12-02
remove unnecessary print statement from sdg_op

### LINKED ISSUE #223: sdg_op component fails with RHELAI 1.3
`sdg_op` with RHEL 1.3 fails at line 76 due to a change in how `read_taxonomy` now works. It now expects an additional parameter `document_output_dir` to be present.   https://github.com/opendatahub-io/ilab-on-ocp/blob/e57915f4f03dbf2f65e33fc9944fb24ac1f28047/sdg/components.py#L76 

This is the error:
```
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

That said, it is unclear to me that line 76 here is doing much. I recommend simply removing it as `read_taxonomy` is called inside of the `generate_data` function anyways.

MichaelClifford:
Closes #223

### DIFF
--- a/pipeline.yaml
+++ b/pipeline.yaml
@@ -1556,8 +1556,7 @@ deploymentSpec:
           \ http_client=custom_http_client\n        )\n    else:\n        client =\
           \ openai.OpenAI(base_url=endpoint, api_key=api_key)\n\n    taxonomy_base\
           \ = \"main\" if repo_branch or (repo_pr and int(repo_pr) > 0) else \"empty\"\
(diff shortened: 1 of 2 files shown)


leseb (approved):

tumido:
Those were originally added by me I think. They come very handy to see the diff between the main branch and PR branch... You know how we had to work on selecting the "right" PR that would generate proper data, here's where you could see it...

Without it you're kinda in blind when seeing a diff on taxonomy repo PR (which you expect to train a model against) but the pipeline is telling you there's no data to train on at the same time.

MichaelClifford:
Closing in favor of #230

PR #228, opened 2024-12-03
Revert "update RHELAI_IMAGE to 1.3"

MichaelClifford:
Reverts opendatahub-io/ilab-on-ocp#222 Reverts opendatahub-io/ilab-on-ocp#225 as simply switching out the RHEL AI images will break the current pipeline. Going to make these change in a larger PR along with fixes needed for RHEL AI 1.3


leseb:
How about we do this in a single PR with 2 revert commits instead?

MichaelClifford:
Closing in favor of a single  revert commit

PR #230, opened 2024-12-04
Pipeline Updates for RHEL AI 1.3

MichaelClifford:
This PR updates the pipeline to use the RHEL AI 1.3 image. 
In addition to updating the image, a couple other updates had to be made:

* set `XDG_CACHE_HOME` to `/tmp` for the `data_processing_task`.
* Update `set_precomputed_skills_data_ratio()` to use a temporary directory when needed in the `sdg_op` component.
* Update PytorchJob manifest to account for changes in available parameters in `main_ds.py`.

PR #231, opened 2024-12-04
Rewrite training component using kubeflow-training library

Shreyanand:
This PR changes the `pytorch_manifest_op` to `pytorch_job_launcher_op` with the following changes:

1. Replace the manifest yaml args, env vars, volumes, volume mounts to kubeflow-training sdk functions 
2. Add master pod train logs to the launcher component so that it is visible in the RHOAI pipeline dashboard
3. Add a flag to delete the pytorch job after completion 

### To do

- [x] The image used in the component needs to be changed to the Python image being created by the RHOAI team.
- [x] Align with RHEL image 1.3
- [x] Change formatting of env vars, volumes, volume mounts

### DIFF
--- a/training/components.py
+++ b/training/components.py
@@ -1,5 +1,5 @@
 # type: ignore
-# pylint: disable=import-outside-toplevel,import-error
+# pylint: disable=import-outside-toplevel,missing-function-docstring
 
 from typing import Optional
 
@@ -8,7 +8,10 @@ from kfp import dsl
 from utils.consts import PYTHON_IMAGE, RHELAI_IMAGE, TOOLBOX_IMAGE
 
 
-@dsl.component(base_image=RHELAI_IMAGE, install_kfp_package=False)
+@dsl.component(
+    base_image=RHELAI_IMAGE,
+    install_kfp_package=False,
+)
 def data_processing_op(
     model_path: str = "/model",
     sdg_path: str = "/data/sdg",
@@ -120,14 +123,16 @@ def knowledge_processed_data_to_artifact_op(
     )
 
 
+# Change base image to the RHOAI python image with kubeflow_training once available
 @dsl.component(base_image=PYTHON_IMAGE, install_kfp_package=False)
-def pytorchjob_manifest_op(
+def pytorch_job_launcher_op(
+    pytorchjob_output_yaml: dsl.Output[dsl.Artifact],
     model_pvc_name: str,
     input_pvc_name: str,
     output_pvc_name: str,
     name_suffix: str,
-    # path_to_model: str,
     phase_num: int,
+    base_image: str,
     nproc_per_node: int = 3,
     nnodes: int = 2,
     num_epochs: int = 2,
@@ -137,26 +142,24 @@ def pytorchjob_manifest_op(
     save_samples: int = 0,
     max_batch_len: int = 20000,
     seed: int = 42,
+    job_timeout: int = 86400,
+    delete_after_done: bool = False,
 ):
-    import inspect
+    import logging
     import os
-    import time
 
-    import kubernetes
-    import urllib3
-    import yaml
+    from kubeflow.training import TrainingClient, models
+    from kubeflow.training.utils import utils as kfto_utils
 
     def list_phase1_final_model():
         model_dir = "/output/phase_1/model/hf_format"
-        models = os.listdir(model_dir)
+        model_list = os.listdir(model_dir)
         newest_idx = max(
             (os.path.getmtime(f"{model_dir}/{model}"), i)
-            for i, model in enumerate(models)
+            for i, model in enumerate(model_list)
         )[-1]
-        newest_model = models[newest_idx]
+        newest_model = model_list[newest_idx]
         return f"{model_dir}/{newest_model}"
-
-    name = f"train-phase-{phase_num}-{name_suffix.rstrip('-sdg')}"
 
     if phase_num == 1:
         path_to_model = "/input_model"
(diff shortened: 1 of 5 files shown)


training/components.py
```diff
@@ -167,111 +169,44 @@ def list_phase1_final_model():
...
+                            --learning_rate={learning_rate} \
+                            --num_warmup_steps={num_warmup_steps} \
+                            --save_samples={save_samples} \
+                            --log_level=INFO \
+                            --max_batch_len={max_batch_len} \
+                            --seed={seed} \
+                            --cpu_offload_optimizer \
+                            --cpu_offload_params \
+                            --distributed_training_framework fsdp \
+                            --is_granite \
```
leseb, line 203: I think this needs to be removed, at least this does not exist in 1.3.

training/components.py
```diff
@@ -296,156 +231,194 @@ def list_phase1_final_model():
...
+    job_template = utils.get_pytorchjob_template(
+        name=name,
+        namespace=namespace,
+        worker_pod_template_spec=worker_pod_template_spec,
+        master_pod_template_spec=master_pod_template_spec,
+        num_workers=nnodes,
+        num_procs_per_worker=nproc_per_node,
+    )
+
+    print(job_template.to_str())
```
leseb, line 359: Is it a debug or something you want to keep?

training/components.py
```diff
@@ -296,156 +231,194 @@ def list_phase1_final_model():
...
+        name="model",
+        persistent_volume_claim=models.V1PersistentVolumeClaimVolumeSource(
+            claim_name=model_pvc_name
+        ),
+    )
+    output_volume = models.V1Volume(
+        name="output",
+        persistent_volume_claim=models.V1PersistentVolumeClaimVolumeSource(
+            claim_name=output_pvc_name
+        ),
```
leseb, line 254: It's a matter of preference but I think it'd be easier to read if you had:

```python
volumes=[
    models.V1Volume(
        name="input-data",
        persistent_volume_claim=models.V1PersistentVolumeClaimVolumeSource(
            claim_name=input_pvc_name
        ),
    ),
    models.V1Volume(
        name="model",
        persistent_volume_claim=models.V1PersistentVolumeClaimVolumeSource(
            claim_name=model_pvc_name
        ),
    )
    ...
    ...
]
```

training/components.py
```diff
@@ -296,156 +231,194 @@ def list_phase1_final_model():
...
+    )
+    input_model_volume_mount = models.V1VolumeMount(
+        mount_path="/input_model", name="model", read_only=True
+    )
+    output_volume_mount_master = models.V1VolumeMount(
+        mount_path="/output", name="output"
+    )
+    output_volume_mount_worker = models.V1VolumeMount(
+        mount_path="/output", name="output", read_only=True
+    )
```
leseb, line 268: ditto

training/components.py
```diff
@@ -296,156 +231,194 @@ def list_phase1_final_model():
...
+
+    # Set env variables
+    nnodes_var = models.V1EnvVar(name="NNODES", value=f"{nnodes}")
+    nproc_per_node_var = models.V1EnvVar(
+        name="NPROC_PER_NODE", value=f"{nproc_per_node}"
+    )
+    xdg_cache_var = models.V1EnvVar(name="XDG_CACHE_HOME", value="/tmp")
+    triton_cache_var = models.V1EnvVar(name="TRITON_CACHE_DIR", value="/tmp")
+    hf_home_var = models.V1EnvVar(name="HF_HOME", value="/tmp")
+    transformers_cache_var = models.V1EnvVar(name="TRANSFORMERS_CACHE", value="/tmp")
```
leseb, line 278: ditto

training/components.py
```diff
@@ -296,156 +231,194 @@ def list_phase1_final_model():
...
+        num_workers=nnodes,
+        num_procs_per_worker=nproc_per_node,
+    )
+
+    print(job_template.to_str())
+
+    # Run the pytorch job
+
+    logging.info(f"Creating PyTorchJob in namespace: {namespace}")
+    training_client.create_job(job_template, namespace=namespace)
```
leseb, line 364: Shouldn't this be part of a try/expect?

MichaelClifford:
@Shreyanand be sure to run `ruff format <file>` to overcome the pre-commit-check errors

training/components.py
```diff
@@ -137,27 +140,26 @@ def pytorchjob_manifest_op(
...
+    import logging
+    from kubeflow.training import TrainingClient
+    from kubeflow.training.utils import utils
+    from kubeflow.training import models
+    import os
 
     def list_phase1_final_model():
         model_dir = "/output/phase_1/model/hf_format"
-        models = os.listdir(model_dir)
+        ilab_models = os.listdir(model_dir)
```
MichaelClifford, line 155: Can we make this `model_list` over `ilab_models`?

training/components.py
```diff
@@ -167,111 +169,44 @@ def list_phase1_final_model():
...
-                  annotations:
-                    sidecar.istio.io/inject: 'false'
-                spec:
-                  containers:
-                    - args:
-                        - |
-                          echo "Running phase {phase_num}"
+    resources_per_worker = {"cpu": "8", "nvidia.com/gpu": nproc_per_node}
+
+    base_image = "quay.io/redhat-et/ilab:shrey"
```
MichaelClifford, line 174: Should this images be `RHELAI_IMAGE` and not this custom image `quay.io/redhat-et/ilab:shrey`?

training/components.py
MichaelClifford: yeah, this will need to be updated according to parameter changes #230.

training/components.py
```diff
@@ -167,111 +169,44 @@ def list_phase1_final_model():
...
+                            --seed={seed} \
+                            --cpu_offload_optimizer \
+                            --cpu_offload_params \
+                            --distributed_training_framework fsdp \
+                            --is_granite \
+                            --checkpoint_at_epoch
+"""
+    ]
+
+    worker_args = [
```
MichaelClifford, line 208: Same for `worker_args`, need to update the main_ds parameters to align with #230

training/components.py
```diff
@@ -137,27 +140,26 @@ def pytorchjob_manifest_op(
...
-    import inspect
-    import os
     import time
-
-    import kubernetes
-    import urllib3
-    import yaml
+    import logging
+    from kubeflow.training import TrainingClient
+    from kubeflow.training.utils import utils
```
MichaelClifford, line 149: nit: slightly confusing when used below since this projects also has a `utils` directory. Can we do something like:

`from kubeflow.training.utils import utils as kfto_utils`

training/components.py
```diff
@@ -296,156 +231,194 @@ def list_phase1_final_model():
...
+        name="NPROC_PER_NODE", value=f"{nproc_per_node}"
+    )
+    xdg_cache_var = models.V1EnvVar(name="XDG_CACHE_HOME", value="/tmp")
+    triton_cache_var = models.V1EnvVar(name="TRITON_CACHE_DIR", value="/tmp")
+    hf_home_var = models.V1EnvVar(name="HF_HOME", value="/tmp")
+    transformers_cache_var = models.V1EnvVar(name="TRANSFORMERS_CACHE", value="/tmp")
+
+    # Get master and worker container specs
+    master_container_spec = utils.get_container_spec(
+        base_image=base_image,
```
MichaelClifford, line 282: This should be `RHELAI_IMAGE` right?

training/components.py
MichaelClifford: I think it would be nice if we wrote this to disc to be inspected.

MichaelClifford (changes requested):

training/components.py
Shreyanand: The `training_client.create_job` has a try/except block for the part where the k8 object is [created](https://github.com/kubeflow/training-operator/blob/master/sdk/python/kubeflow/training/api/training_client.py#L517). And this deals with all the Exceptions. 
I could add another try/except for all Exceptions in our script for the parts not covered in the above try/except block. WDYT?

leseb: Yes, if there is a good exception to catch let's do that. Thanks

Shreyanand:
> @Shreyanand be sure to run `ruff format <file>` to overcome the pre-commit-check errors

@MichaelClifford I did run it and it doesn't change any file but somehow the pre-commit test fails 🤔 
Screenshot 2024-12-10 at 3 11 29 PM

leseb:
> > @Shreyanand be sure to run `ruff format <file>` to overcome the pre-commit-check errors
> 
> @MichaelClifford I did run it and it doesn't change any file but somehow the pre-commit test fails 🤔 Screenshot 2024-12-10 at 3 11 29 PM

It's the import that are failing so you need to run: `ruff check --select I --fix .`, thanks!

training/components.py
```diff
@@ -292,154 +230,175 @@ def list_phase1_final_model():
...
+    logging.info(f"Monitoring job until status is any of {expected_conditions}.")
+
+    def wait_for_job_get_logs(
+        name: str,
+        namespace: str = None,
+        job_kind: str = None,
+        expected_conditions: list = ["Succeeded"],
+        wait_timeout: int = 600,
+        polling_interval: int = 15,
+        timeout: int = 1000,
```
tumido, line 356: Can you please explain this function to me a bit? I don't follow.. Can't be a simple call:

```
training_client.get_jobs_log(follow=True, ...)
training_client.get_job_conditions()
```

Based on the [`get_jobs_log()` sources](https://github.com/kubeflow/training-operator/blob/374e88af2d28d49e1c40008238f6e039b5677177/sdk/python/kubeflow/training/api/training_client.py#L1210-L1255) the `follow=True` should be blocking and stream all the logs until all the pods finish.

Is there anything wrong with that approach? I may have overlooked it...

training/components.py
```diff
@@ -292,154 +230,175 @@ def list_phase1_final_model():
...
+        models.V1VolumeMount(
+            mount_path="/input_data", name="input-data", read_only=True
+        ),
+        models.V1VolumeMount(mount_path="/input_model", name="model", read_only=True),
+    ]
+    volume_mounts_master = volume_mounts_common.append(
+        models.V1VolumeMount(mount_path="/output", name="output")
+    )
+    volume_mounts_worker = volume_mounts_common.append(
+        models.V1VolumeMount(mount_path="/output", name="output", read_only=True)
```
tumido, line 269: This doesn't work. `list.append()` returns `None` in Python and modifies the original array:


```python
>>> a = []
>>> b = a.append("foo")
>>> a
['foo']
>>> b
# Nothing is returned
```

tumido (changes requested):

training/components.py
tumido: Yes, the value here should be the same as `RHELAI_IMAGE`, but due to KFP limitations, the Pythonic components are hermetic and you can't reference external constanstants here...

https://www.kubeflow.org/docs/components/pipelines/user-guides/components/lightweight-python-components/#python-function-requirements

I think we can get around that if we pass the image for `PytorchJob` as component parameter which defaults to the `RHELAI_IMAGE` constant.

training/components.py
Shreyanand: I tried `follow=True` flag but it doesn't block the launcher container from finishing. I also tried [wait_for_job_condition](https://github.com/kubeflow/training-operator/blob/374e88af2d28d49e1c40008238f6e039b5677177/sdk/python/kubeflow/training/api/training_client.py#L888) which blocks the launcher but doesn't stream logs. If I use both of them, the `training_client.get_jobs_log(follow=True, ...)` prints the log stream to stdout but if we want to see it in pod logs and on the UI logs for the launcher, we have to use `logging.info()` to show them. 

The above function is similar to `wait_for_job_condition` but adds the get pod logs while waiting feature. It's good that you bring it up, maybe there can be a better way to do all of this.

training/components.py
Shreyanand: Fixed, thanks 👍

training/components.py
Shreyanand: Added as a yaml output artifact of the launcher component.

training/components.py
```diff
@@ -167,108 +170,43 @@ def list_phase1_final_model():
...
-                  annotations:
-                    sidecar.istio.io/inject: 'false'
-                spec:
-                  containers:
-                    - args:
-                        - |
-                          echo "Running phase {phase_num}"
+    resources_per_worker = {"nvidia.com/gpu": nproc_per_node}
+
+    base_image = "registry.stage.redhat.io/rhelai1/instructlab-nvidia-rhel9:1.3.1"
```
tumido, line 175: Can we pass this as a task param? That way we should be able to do https://github.com/opendatahub-io/ilab-on-ocp/pull/231#discussion_r1880444728

```python
def pytorch_job_launcher_op(base_image=RHELAI_IMAGE,...
```

training/components.py
tumido: Just curious.. this you also try the `wait_for_job_condition` with `callback` param?

https://github.com/kubeflow/training-operator/blob/374e88af2d28d49e1c40008238f6e039b5677177/sdk/python/kubeflow/training/api/training_client.py#L896

It should allow us to get the logs via a callback function on each polling interval.

training/components.py
Shreyanand: @tumido I tried doing that but the pipelines [errors out](https://rhods-dashboard-redhat-ods-applications.apps.ocp-beta-test.nerc.mghpcc.org/experiments/ilab/5d5a1755-0df4-4a08-9a46-85b1582be2f5/runs/93655405-47e0-4c38-ba61-908b85e199c7). Basically, the constant is not available for the function defaults as well. One option could be to not set a default value for the `base_image` parameter and import the constant `RHELAI_IMAGE` in pipeline.py and pass that as a paramter while calling the `pytorch_job_launcher_op` component. WDYT?

tumido: For the context few screenshots from @Shreyanand's run:

image
image
image
image

That's very interesting. The schema parses the param value properly, passes it as input yet the execution fails on unknown constant.. Looks like a KFP bug... Very weird behavior.


> One option could be to not set a default value for the base_image parameter and import the constant RHELAI_IMAGE in pipeline.py and pass that as a paramter while calling the pytorch_job_launcher_op component. WDYT?

Knowing the above I'm not convinced it would work, but please try that 🙂

Another option/workaround would be to read that value from env variable. 🙂 And this one should work. Cause KFP task runtime is not involved in parsing env variables, that goes directly from the schema.

Shreyanand: > Knowing the above I'm not convinced it would work, but please try that 🙂

So, it seems to be working, we just have to import the image name in the `pipeline.py` file. Weird behaviour indeed. I'll test it again and confirm with a commit.

training/components.py
Shreyanand: I did not try that. Turns out, the callback function option with the get_job_logs(..., follow=True) does work for us. The only caveat is that the callback function can only have the job object as input whereas get_job_logs takes in the job name as an input. So, as a workaround I had to create another function to bridge that mismatch. I'll add that shortly, but awesome suggestion, it significantly reduces the unnecessary code 💯

training/components.py
Shreyanand: Added

leseb (approved):
Nice job!

tumido (commented):
Just a nit. Can we get this one formatted better:
https://github.com/opendatahub-io/ilab-on-ocp/pull/231/files#r1887046992

Otherwise LGTM 🙂

tumido (approved):
/lgtm
/approve

Shreyanand:
@tumido @HumairAK @MichaelClifford 
I rebased this over the latest image fixes in #255 and a completed test run is here: https://rhods-dashboard-redhat-ods-applications.apps.ocp-beta-test.nerc.mghpcc.org/experiments/ilab/5d5a1755-0df4-4a08-9a46-85b1582be2f5/runs/e8c6ab40-1944-4b64-9b21-7be6a00b910f

PR #233, opened 2024-12-09
chore: ensure pipeline params match ilab defaults

### LINKED ISSUE #211: Re-evaluate default param values for KFP
Let's make sure default parameters for KFP do match requirements and make sense in generic environment.

tumido:
Resolves #211

## Description
Re-align defaults we use in the pipeline with what InstructLab declares as defaults in https://github.com/instructlab/instructlab/blob/main/tests/testdata/default_config.yaml

Each param default now features a reference to the InstructLab default config.

**WIP**. Some params require additional discussion. I've marked them as `FIXME:`. These params have no basis in the default config and we should discuss if they need to be surfaced, if they require further justification for its default value or they are good as is. They can totally be good as is, I just wanted to raise awareness about them.

### DIFF
--- a/pipeline.py
+++ b/pipeline.py
@@ -23,31 +23,10 @@ JUDGE_SECRET = "judge-server"
 MOCKED_STAGES = ["sdg", "train", "eval"]
 PIPELINE_FILE_NAME = "pipeline.yaml"
 IMPORTER_PIPELINE_FILE_NAME = "importer-pipeline.yaml"
-SDG_PIPELINE = "simple"
 IMAGE_PULL_SECRET = "redhat-et-ilab-botty-pull-secret"
 STANDALONE_TEMPLATE_FILE_NAME = "standalone.tpl"
 GENERATED_STANDALONE_FILE_NAME = "standalone.py"
 DEFAULT_REPO_URL = "https://github.com/instructlab/taxonomy.git"
-
-# eval args
-FEW_SHOTS = 5
-# BATCH_SIZE can also be an int, for example "8" is converted to an int in eval/final
-BATCH_SIZE = "auto"
-MAX_WORKERS = "auto"
-MERGE_SYSTEM_USER_MESSAGE = False
-
-# training args
-NUM_EPOCHS_PHASE_1 = 2
-NUM_EPOCHS_PHASE_2 = 2
-EFFECTIVE_BATCH_SIZE_PHASE_1 = 3840
-EFFECTIVE_BATCH_SIZE_PHASE_2 = 3840
-LEARNING_RATE_PHASE_1 = 1e-4
-LEARNING_RATE_PHASE_2 = 1e-4
-NUM_WARMUP_STEPS_PHASE_1 = 100
-NUM_WARMUP_STEPS_PHASE_2 = 100
-SAVE_SAMPLES = 0
-MAX_BATCH_LEN = 20000
-SEED = 42
 
 
 def ilab_pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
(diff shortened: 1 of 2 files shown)


pipeline.py
```diff
@@ -112,32 +91,32 @@ def pipeline(
             int
         ] = None,  # FIXME: https://issues.redhat.com/browse/RHOAIRFE-467
         sdg_base_model: str = "s3://<BUCKET>/<PATH_TO_MODEL>",
-        sdg_scale_factor: int = 2,  # Renamed upstream https://github.com/instructlab/instructlab/blob/f7d40f6ed5112d59132dd832bd332fa6fbbe7010/src/instructlab/configuration.py#L279-L290
-        sdg_pipeline: str = SDG_PIPELINE,
-        sdg_max_batch_len: int = MAX_BATCH_LEN,
-        sdg_sample_size: float = 1.0,
+        sdg_scale_factor: int = 30,  # https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72/tests/testdata/default_config.yaml#L125
```
leseb, line 94: I think pointing to [d68c5a12026e08092bd4171a77667b09f7c36f72](https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72) is "risky". Why this hash in particular? I'd prefer to use a tag at best.

pipeline.py
```diff
@@ -112,32 +91,32 @@ def pipeline(
...
+        sdg_pipeline: str = "full",  # https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72/tests/testdata/default_config.yaml#L122
+        sdg_max_batch_len: int = 5000,  # https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72/tests/testdata/default_config.yaml#L334
+        sdg_sample_size: float = 1.0,  # FIXME: Not present in default config. Need justification?
```
leseb, line 97: You can point to https://github.com/instructlab/sdg/pull/432 for further details.

pipeline.py
tumido: That was just the most recent commit at `main` at the time me pulling the data (and still is as of now). Unless you lock a link to specific commit, it's not a permanent reference. Lines can shift in time and mapping to a tag/branch is a moving target. So linking to line e.g. 123 would point to a different property over time...

pipeline.py
```diff
@@ -112,32 +91,32 @@ def pipeline(
...
-        train_effective_batch_size_phase_1: int = EFFECTIVE_BATCH_SIZE_PHASE_1,
-        train_effective_batch_size_phase_2: int = EFFECTIVE_BATCH_SIZE_PHASE_2,
-        train_learning_rate_phase_1: float = LEARNING_RATE_PHASE_1,
-        train_learning_rate_phase_2: float = LEARNING_RATE_PHASE_2,
-        train_num_warmup_steps_phase_1: int = NUM_WARMUP_STEPS_PHASE_1,
-        train_num_warmup_steps_phase_2: int = NUM_WARMUP_STEPS_PHASE_2,
-        train_save_samples: int = SAVE_SAMPLES,
-        train_max_batch_len: int = MAX_BATCH_LEN,
-        train_seed: int = SEED,
+        train_nproc_per_node: int = 3,  # FIXME: Not present in default config. Need justification?
```
leseb, line 99: https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72/tests/testdata/default_config.yaml#L345

pipeline.py
```diff
@@ -112,32 +91,32 @@ def pipeline(
...
+        train_nnodes: int = 2,  # FIXME: Not present in default config. Need justification?
```
leseb, line 100: ilab does not do any distributed training across nodes so this is normal such flag does not exist.

tumido: The question is, do we have a justification for the defaults we've selected here? Do we need it? Is 2*3 the best default config we can produce? 🙂

pipeline.py
tumido: Same question as https://github.com/opendatahub-io/ilab-on-ocp/pull/233#discussion_r1876127054 basically...

I know ILab is 1*1 on GPUs by default. Should that be the default in a distributed environment we're creating this for?

pipeline.py
```diff
@@ -112,32 +91,32 @@ def pipeline(
...
+        train_num_epochs_phase_2: int = 10,  # https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72/tests/testdata/default_config.yaml#L377
+        train_effective_batch_size_phase_1: int = 128,  # https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72/tests/testdata/default_config.yaml#L357
+        train_effective_batch_size_phase_2: int = 3840,  # https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72/tests/testdata/default_config.yaml#L371
+        train_learning_rate_phase_1: float = 2e-05,  # https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72/tests/testdata/default_config.yaml#L360
+        train_learning_rate_phase_2: float = 6e-06,  # https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72/tests/testdata/default_config.yaml#L374
+        train_num_warmup_steps_phase_1: int = 100,  # FIXME: Not present in default config. Need justification?
```
leseb, line 107: Not exposed by `ilab` CLI but present in the training lib https://github.com/instructlab/training/blob/main/src/instructlab/training/main_ds.py#L874

pipeline.py
```diff
@@ -112,32 +91,32 @@ def pipeline(
...
+        train_num_warmup_steps_phase_2: int = 100,  # FIXME: Not present in default config. Need justification?
+        train_save_samples: int = 250000,  # https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72/tests/testdata/default_config.yaml#L393
+        train_max_batch_len: int = 5000,  # https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72/tests/testdata/default_config.yaml#L334
+        train_seed: int = 42,  # FIXME: Not present in default config. Need justification?
```
leseb, line 111: Not exposed by ilab CLI but present in the training lib https://github.com/instructlab/training/blob/main/src/instructlab/training/main_ds.py#L901

pipeline.py
```diff
@@ -112,32 +91,32 @@ def pipeline(
...
         # MT Bench
-        mt_bench_max_workers: str = MAX_WORKERS,
-        mt_bench_merge_system_user_message: bool = MERGE_SYSTEM_USER_MESSAGE,
+        mt_bench_max_workers: str = "auto",  # https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72/tests/testdata/default_config.yaml#L74
+        mt_bench_merge_system_user_message: bool = False,  # FIXME: Not present in default config. Need justification?
```
leseb, line 114: CLI flag: https://github.com/instructlab/instructlab/blob/main/src/instructlab/model/evaluate.py#L474 not expose in config

pipeline.py
```diff
@@ -112,32 +91,32 @@ def pipeline(
...
         # Final evaluation
-        final_eval_max_workers: str = MAX_WORKERS,
-        final_eval_few_shots: int = FEW_SHOTS,
-        final_eval_batch_size: str = BATCH_SIZE,
-        final_eval_merge_system_user_message: bool = MERGE_SYSTEM_USER_MESSAGE,
+        final_eval_max_workers: str = "auto",  # FIXME: Not present in default config. Need justification?
+        final_eval_few_shots: int = 5,  # https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72/tests/testdata/default_config.yaml#L56
+        final_eval_batch_size: str = "auto",  # https://github.com/instructlab/instructlab/blob/d68c5a12026e08092bd4171a77667b09f7c36f72/tests/testdata/default_config.yaml#L52
+        final_eval_merge_system_user_message: bool = False,  # FIXME: Not present in default config. Need justification?
```
leseb, line 119: CLI flag: https://github.com/instructlab/instructlab/blob/main/src/instructlab/model/evaluate.py#L474 not expose in config

pipeline.py
MichaelClifford: I think we should either do 2*1 or 2*2 as default.  

2*1 would be the absolute minimum for mulit-pod distribution.
2*2 would be the minimum for multi-gpu, multi-pod distribution.

I vote for 2*2 so the default is showing the full capabilities of the  system, but happy to hear other opinions.

pipeline.py
leseb: Ok, but what's the issue with tag + line number? I think it's more natural and easier to understand since the tag will point to a given version. Thanks!

pipeline.py
tumido: Sure I can. However I don't understand it 😄 Can you please help me understand how that PR explains that `1.0` is a good default value? The PR doesn't even mention `1.0` being a valid value...

pipeline.py
leseb: Definitely not 1. While this file is good, I don't think we should default from all the values. I'd pick a reference architecture that fits those https://github.com/instructlab/instructlab/blob/main/src/instructlab/configuration.py#L621-L915. Then we apply the value from the profile that matches our ref arch.

pipeline.py
tumido: Well.. tag is also moving target, but guess this is not abused that often. 😄 I think that is a good compromise. 🙂

pipeline.py
tumido: Ok.. so what if I leave those `FIXME` in place for the time being and explain that we're waiting for a reference architecture? WDYT? Would that make sense?

tumido: I can combine that with what @MichaelClifford says above.. change that 2x2 + add a note the RHOAI team needs to provide their justification or change it. 🙂

tumido: Updated both to be 2x2 and added a note. WDYT now? 🙂

pipeline.py
leseb: Sorry I don't mean to argue over this but I don't understand the "tag is also moving target", a tag is applied to a commit and considered immutable. While it’s technically possible to move or delete a tag, this is quite rare and generally discouraged.

Ok great for using a tag + line number which is a pretty stable and understandable reference :-)

pipeline.py
leseb: Yes, sounds good, thanks!

tumido: Thank you for having the link at your disposal. 🙂 Really. 🙂  It's gonna make the handover so much less painful. 🙌

pipeline.py
```diff
@@ -112,32 +91,32 @@ def pipeline(
...
-        mt_bench_max_workers: str = MAX_WORKERS,
-        mt_bench_merge_system_user_message: bool = MERGE_SYSTEM_USER_MESSAGE,
+        mt_bench_max_workers: str = "auto",  # https://github.com/instructlab/instructlab/blob/v0.21.2/tests/testdata/default_config.yaml#L74
+        mt_bench_merge_system_user_message: bool = False,  # https://github.com/instructlab/instructlab/blob/v0.21.2/src/instructlab/model/evaluate.py#L474
         # Final evaluation
-        final_eval_max_workers: str = MAX_WORKERS,
-        final_eval_few_shots: int = FEW_SHOTS,
-        final_eval_batch_size: str = BATCH_SIZE,
-        final_eval_merge_system_user_message: bool = MERGE_SYSTEM_USER_MESSAGE,
+        final_eval_max_workers: str = "auto",  # FIXME: Not present in default config. Need justification?
```
leseb, line 116: https://github.com/instructlab/instructlab/blob/v0.21.2/tests/testdata/default_config.yaml#L74 this is used for both mt_bench and mt_bench_branch (AKA final eval :p)

pipeline.py
```diff
@@ -558,7 +537,12 @@ def run(mock, experiment, run_name, param):
     dev_arguments = {
         "k8s_storage_class_name": "nfs-csi",
         "sdg_base_model": "s3://ilab-pipeline-b1d4c2b1-ab00-4e7f-b985-697bda3df385/instructlab-base-importer/648f36d0-e3f0-43b8-8adb-530576beb675/ilab-importer-op/model/granite-7b-starter",
-        "train_nproc_per_node": 2,
+        "train_num_epochs_phase_1": 2,
```
leseb, line 540:
```suggestion
        "train_num_epochs_phase_1": 2,
        "sdg_sample_size": 0.0002,
```

leseb: cc @MichaelClifford

pipeline.py
tumido: Should we also merge it into a single param then? WDYT @MichaelClifford ? 🙂 

Right now we have 2 separate params:
- `final_eval_max_workers` - used in MT Bench Branch
- `mt_bench_max_workers` - used in MT Bench

leseb (commented):
Do we need to update the README as well?

pipeline.py
MichaelClifford: I think we separated this out in an earlier PR since we didn't want to assume that two different processes will always need the same a parameter. However, I do see that in the default config they do not differentiate. 🤔 I'm fine keeping it as 2 separate params, but would also not care if we make it one. So its up to you @tumido 😄

pipeline.py
MichaelClifford: 👍 yep

MichaelClifford (approved):
LGTM!

pipeline.py
tumido: I'll keep it as is. If RHOAI/RHEL AI sees the need to unify these they always can. 🙂

leseb:
> Do we need to update the README as well?

@tumido do we need any changes in https://github.com/opendatahub-io/ilab-on-ocp?tab=readme-ov-file#available-pipeline-parameters?

tumido:
I don't think so. No param was removed or added, no description has changed... So in regard to the content we already have in the README, no change is required.

However I can extend the README with default values and references, if we want to link it from there as well. 🤷 I don't have a preference there. What would you do? 🙂 @MichaelClifford WDYT?

leseb:
> I don't think so. No param was removed or added, no description has changed... So in regard to the content we already have in the README, no change is required.
> 
> However I can extend the README with default values and references, if we want to link it from there as well. 🤷 I don't have a preference there. What would you do @leseb ? 🙂 @MichaelClifford WDYT?

Maybe as a followup if you feel like it. Thanks!

leseb (approved):

MichaelClifford:
> >I don't think so. No param was removed or added, no description has changed... So in regard to the content we already have in the README, no change is required.
> 
> However I can extend the README with default values and references, if we want to link it from there as well. 🤷 I don't have a preference there. What would you do @leseb ? 🙂 @MichaelClifford WDYT?
> 
> Maybe as a followup if you feel like it. Thanks!

Agree let's do it in a follow up, if at all 😄

MichaelClifford (approved):

PR #239, opened 2024-12-17
set caching to False for final_eval_task

### LINKED ISSUE #238: Missing `.set_caching_options(False)` for final eval step
Missing `.set_caching_options(False)` for final eval step. This can result in the final eval not running if it has successfully completed in an earlier run. We want to ensure that this step always runs, so we need to add `fina_eval_task.set_caching_option(False)` to `pipeline.py`

MichaelClifford:
closes #238


leseb (approved):
Makes sense.

PR #241, opened 2024-12-18
Add additional step to pipeline to generate a metrics report

MichaelClifford:
This PR adds a new step to the end of the pipeline,`generate_metrics_report_op`. The purpose of this step is to create a number of kfp `Mertic` artifacts that can help users easily compare the performance between different runs using the `compare runs` feature of the Data Science Pipeline UI.

### DIFF
--- a/eval/final/components.py
+++ b/eval/final/components.py
@@ -1,9 +1,9 @@
 # type: ignore
 # pylint: disable=no-value-for-parameter,import-outside-toplevel,import-error
 
-from kfp.dsl import Artifact, Output, component
+from kfp.dsl import Artifact, Input, Metrics, Output, component
 
-from utils.consts import RHELAI_IMAGE
+from utils.consts import PYTHON_IMAGE, RHELAI_IMAGE
 
 
 @component(base_image=RHELAI_IMAGE, install_kfp_package=False)
@@ -480,3 +480,36 @@ def run_final_eval_op(
 
     with open(mt_bench_branch_output.path, "w", encoding="utf-8") as f:
         json.dump(mt_bench_branch_data, f, indent=4)
+
+
+@component(base_image=PYTHON_IMAGE, install_kfp_package=False)
+def generate_metrics_report_op(
+    mmlu_branch_output: Input[Artifact],
+    mt_bench_branch_output: Input[Artifact],
+    mt_bench_output: Input[Artifact],
+    metrics: Output[Metrics],
+):
+    import ast
+    import json
+
+    with open(mt_bench_output.path, "r") as f:
+        mt_bench_data = f.read()
+    mt_bench_data = ast.literal_eval(mt_bench_data)[0]
+
+    metrics.log_metric("mt_bench_best_model", mt_bench_data["model"])
+    metrics.log_metric("mt_bench_best_score", mt_bench_data["overall_score"])
+    metrics.log_metric("mt_bench_best_model_error_rate", mt_bench_data["error_rate"])
+
+    with open(mt_bench_branch_output.path, "r") as f:
+        mt_bench_branch_data = json.loads(f.read())
+
+    metrics.log_metric("mt_bench_branch_score", mt_bench_branch_data["overall_score"])
+    metrics.log_metric(
+        "mt_bench_branch_base_score", mt_bench_branch_data["base_overall_score"]
+    )
+
+    with open(mmlu_branch_output.path, "r") as f:
+        mmlu_branch_data = json.loads(f.read())
+
+    metrics.log_metric("mmlu_branch_score", mmlu_branch_data["model_score"])
+    metrics.log_metric("mmlu_branch_base_score", mmlu_branch_data["base_model_score"])
--- a/pipeline.py
+++ b/pipeline.py
@@ -73,7 +73,7 @@ def ilab_pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
         )
 
     # Imports for evaluation
-    from eval.final import run_final_eval_op
+    from eval.final import generate_metrics_report_op, run_final_eval_op
     from eval.mt_bench import run_mt_bench_op
 
     @dsl.pipeline(
@@ -427,6 +427,14 @@ def ilab_pipeline_wrapper(mock: List[Literal[MOCKED_STAGES]]):
 
         model_pvc_delete_task = DeletePVC(pvc_name=model_pvc_task.output)
         model_pvc_delete_task.after(final_eval_task)
+
+        generate_metrics_report_task = generate_metrics_report_op(
+            mmlu_branch_output=final_eval_task.outputs["mmlu_branch_output"],
+            mt_bench_branch_output=final_eval_task.outputs["mt_bench_branch_output"],
+            mt_bench_output=output_mt_bench_task.outputs["mt_bench_output"],
+        )
+        generate_metrics_report_task.after(output_mt_bench_task)
+        generate_metrics_report_task.set_caching_options(False)
 
         return
 
--- a/eval/final/__init__.py
+++ b/eval/final/__init__.py
@@ -1,5 +1,5 @@
-from .components import run_final_eval_op
+from .components import generate_metrics_report_op, run_final_eval_op
 
 # from . import faked
 
-__all__ = ["run_final_eval_op"]
+__all__ = ["run_final_eval_op", "generate_metrics_report_op"]
--- a/pipeline.yaml
+++ b/pipeline.yaml
@@ -30,6 +30,8 @@
 #    train_num_warmup_steps_phase_2: int [Default: 1000.0]
 #    train_save_samples: int [Default: 250000.0]
 #    train_seed: int [Default: 42.0]
+# Outputs:
+#    generate-metrics-report-op-metrics: system.Metrics
 components:
   comp-createpvc:
     executorLabel: exec-createpvc
@@ -266,6 +268,28 @@ components:
           description: Name of the PVC to delete. Supports passing a runtime-generated
             name, such as a name provided by ``kubernetes.CreatePvcOp().outputs['name']``.
           parameterType: STRING
+  comp-generate-metrics-report-op:
+    executorLabel: exec-generate-metrics-report-op
+    inputDefinitions:
+      artifacts:
+        mmlu_branch_output:
+          artifactType:
+            schemaTitle: system.Artifact
+            schemaVersion: 0.0.1
+        mt_bench_branch_output:
+          artifactType:
+            schemaTitle: system.Artifact
+            schemaVersion: 0.0.1
+        mt_bench_output:
+          artifactType:
+            schemaTitle: system.Artifact
+            schemaVersion: 0.0.1
+    outputDefinitions:
+      artifacts:
+        metrics:
+          artifactType:
+            schemaTitle: system.Metrics
+            schemaVersion: 0.0.1
   comp-git-clone-op:
     executorLabel: exec-git-clone-op
     inputDefinitions:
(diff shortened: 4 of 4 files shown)


eval/final/components.py
```diff
@@ -480,3 +480,36 @@ def find_node_dataset_directories(base_dir: str):
 
     with open(mt_bench_branch_output.path, "w", encoding="utf-8") as f:
         json.dump(mt_bench_branch_data, f, indent=4)
+
+
+@component(base_image=PYTHON_IMAGE, install_kfp_package=False)
+def generate_metrics_report_op(
+    mmlu_branch_output: Input[Artifact],
+    mt_bench_branch_output: Input[Artifact],
+    mt_bench_output: Input[Artifact],
```
tumido, line 489: WDYT about making these inputs being read from the PVC instead of them being input artifact (as the other steps do)? It would make things faster (and more uniform).

Other option would be to emit those metrics directly from `final_eval_task` and `output_mt_bench_task` steps. I think that would be the best option, WDYT?

MichaelClifford: Emitting them from `final_eval_task` and `output_mt_bench_task` was my first approach, but it has some bad consequences for the UX, specifically metrics from each step get rendered in their own sub-column per run column. So it becomes difficult to read, and if we want to add more metrics from other steps the problem will get worse. So I think keeping a single step that collects all the metrics we want to report out is the best approach for now, as it keeps everything ins a single column and makes comparing run metrics easier.

MichaelClifford: >WDYT about making these inputs being read from the PVC instead of them being input artifact (as the other steps do)? It would make things faster (and more uniform).

Great idea 😄  I will make that change, Thanks!

MichaelClifford:
@tumido updated the PR to use PVC's instead of interim output artifacts.

pipeline.py
```diff
@@ -452,17 +452,29 @@ def pipeline(
...
+            task=generate_metrics_report_task,
+            pvc_name=output_pvc_task.output,
+            mount_path="/output",
+        )
+
+        output_pvc_delete_task = DeletePVC(pvc_name=output_pvc_task.output)
+        output_pvc_delete_task.after(
+            output_model_task,
+            output_mt_bench_task,
+            final_eval_task,
```
tumido, line 474:
```suggestion
```

No need for these 2 dependencies, since the `generate_metrics_report_task` already depends on it here https://github.com/opendatahub-io/ilab-on-ocp/pull/241/files#diff-6b92e3e77956f8020482ca91889a24f3411c9a4854f2aaaac30ee504ef26ede4R462

eval/final/components.py
```diff
@@ -470,5 +475,41 @@ def find_node_dataset_directories(base_dir: str):
...
+@component(base_image=PYTHON_IMAGE, install_kfp_package=False)
+def generate_metrics_report_op(
+    metrics: Output[Metrics],
+):
+    import ast
+    import json
+
+    with open("/output/mt_bench_data.json", "r") as f:
+        mt_bench_data = f.read()
+    mt_bench_data = ast.literal_eval(mt_bench_data)[0]
```
tumido, line 497: Just a curious question. Why is `json.loads()` not enough here? Why do we need `eval`?

eval/mt_bench/components.py
```diff
@@ -188,14 +188,16 @@ def shutdown_vllm(process: subprocess.Popen, timeout: int = 20):
         scores[model_path] = overall_score
 
     with open(output_path, "w", encoding="utf-8") as f:
-        json.dump(all_mt_bench_data, f, indent=4)
+        for data in all_mt_bench_data:
+            json.dump(data, f)
+            f.write("\n")
```
tumido, line 193: Can't this be a JSON instead of JSONL? It's a JSON Array in the end. A file containing `[{...},{...}]` is a valid JSON, isn't it?

I still don't understand the parsing issue, if it was a JSON in the first place...

eval/mt_bench/components.py
```diff
@@ -188,14 +188,16 @@ def shutdown_vllm(process: subprocess.Popen, timeout: int = 20):
...
 
     outputs = NamedTuple("outputs", best_model=str, best_score=float)
     best_model = max(scores, key=scores.get)
     best_score = scores[best_model]
-    if best_score_file:
-        with open(best_score_file, "w", encoding="utf-8") as f:
-            json.dump({"best_model": best_model, "best_score": best_score}, f, indent=4)
+
+    with open(best_score_file, "w", encoding="utf-8") as f:
+        json.dump({"best_model": best_model, "best_score": best_score}, f, indent=4)
```
tumido, line 200: If we are creating a new file, would it be worth uploading it as an artifact as well? (like we do with the other file via `pvc_to_mt_bench_op` task)

MichaelClifford: We could. Seems like its somewhat duplicative information though. Might make sense to just insert this information into the top of the `output/mt_bench_data.json` instead and keep it all in a single file. WDYT?

eval/mt_bench/components.py
MichaelClifford: You are right. I re-evaluated this step just using json instead of jsonl and it seems to work as expected. 🙃 Will revert this change.

eval/final/components.py
```diff
@@ -470,5 +475,38 @@ def find_node_dataset_directories(base_dir: str):
...
         json.dump(mt_bench_branch_data, f, indent=4)
+
+
+@component(base_image=PYTHON_IMAGE, install_kfp_package=False)
+def generate_metrics_report_op(
+    metrics: Output[Metrics],
+):
+    import json
+
+    with open("/output/mt_bench_best_data.json", "r") as f:
```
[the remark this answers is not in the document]
MichaelClifford: I removed this all together in favor of keeping all mt_bench info in the same file `/output/mt_bench_data.json`.

eval/final/components.py
```diff
@@ -470,5 +475,38 @@ def find_node_dataset_directories(base_dir: str):
...
+    with open(
+        f"{mt_bench_branch_output_path}/mt_bench_branch_data.json",
+        "w",
+        encoding="utf-8",
+    ) as f:
         json.dump(mt_bench_branch_data, f, indent=4)
+
+
+@component(base_image=PYTHON_IMAGE, install_kfp_package=False)
+def generate_metrics_report_op(
```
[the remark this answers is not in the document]
MichaelClifford: Good suggestion 😄  Done!

eval/mt_bench/components.py
MichaelClifford: I went ahead and consolidated all mt_bench info into `mt_bench_data.json` so no longer any need to mange this extra file.

eval/final/components.py
```diff
@@ -20,10 +18,13 @@ def run_final_eval_op(
     candidate_model: str = None,
     taxonomy_path: str = "/input/taxonomy",
     sdg_path: str = "/input/sdg",
+    mmlu_branch_output_path: str = "/output/mmlu_branch",
+    mt_bench_branch_output_path: str = "/output/mt_bench_branch",
```
tumido, line 22: This change means we no longer output this file(s) as artifacts. We only store the data on PVC and generate metrics from it. But we no longer upload it as `Output[Artifact]` Is this intentionally replaced with the metrics?

If yes, we should also remove `output_mt_bench_task`
https://github.com/opendatahub-io/ilab-on-ocp/pull/241/files#diff-6b92e3e77956f8020482ca91889a24f3411c9a4854f2aaaac30ee504ef26ede4R443-R445 for consistency (uploading `mt_bench` only JSON and omitting `mmlu_branch` and `mt_bench_branch` makes no sense).

If no, we should introduce similar upload step(s) as `output_mt_bench_task` for `mmlu_branch` and `mt_bench_branch` or join those with `output_mt_bench_task` into a single step...

eval/final/components.py
```diff
@@ -1,15 +1,13 @@
 # type: ignore
 # pylint: disable=no-value-for-parameter,import-outside-toplevel,import-error
 
-from kfp.dsl import Artifact, Output, component
+from kfp.dsl import Artifact, Input, Metrics, Output, component
```
tumido, line 4: See below, if you intend to remove `Output[Artifact]`, the `Artifact` and `Output` imports are no longer needed here.

eval/mt_bench/components.py
```diff
@@ -187,15 +186,13 @@ def shutdown_vllm(process: subprocess.Popen, timeout: int = 20):
...
-    with open(output_path, "w", encoding="utf-8") as f:
-        json.dump(all_mt_bench_data, f, indent=4)
-
     outputs = NamedTuple("outputs", best_model=str, best_score=float)
     best_model = max(scores, key=scores.get)
     best_score = scores[best_model]
-    if best_score_file:
-        with open(best_score_file, "w", encoding="utf-8") as f:
-            json.dump({"best_model": best_model, "best_score": best_score}, f, indent=4)
+    all_mt_bench_data.insert(0, {"best_model": best_model, "best_score": best_score})
```
leseb, line 192: Though for later: I think the data layout is not ideal. Forcing the index 0 can be tricky and lead to errors later. Now we have something like this:

```
[
    {
        "best_model": best_model,
        "best_score": best_score
    },
    {
        "report_title": "SKILLS EVALUATION REPORT",
        "model": "model",
        "judge_model": "judge_model",
        "overall_score": overall_score,
        "turn_scores": turn_scores,
        "qa_scores": qa_scores,
        "error_rate": error_rate
    }
]
```

I think having a layout like would be easier:

```
{
    "best_model": best_model,
    "best_score": best_score,
    "reports": [
        {
            "report_title": "SKILLS EVALUATION REPORT",
            "model": model_path,
            "judge_model": judge_model_name,
            "overall_score": overall_score,
            "turn_scores": turn_scores,
            "qa_scores": qa_pairs,
            "error_rate": error_rate,
        },
        {
            "report_title": "SKILLS EVALUATION REPORT",
            "model": model_path,
            "judge_model": judge_model_name,
            "overall_score": overall_score,
            "turn_scores": turn_scores,
            "qa_scores": qa_pairs,
            "error_rate": error_rate,
        }
    ]
}
```

So that we can go `data.get("best_model")`. This should be a later refactor though.

eval/final/components.py
```diff
@@ -320,14 +321,20 @@ def find_node_dataset_directories(base_dir: str):
...
             "summary": summary,
         }
 
-        with open(mmlu_branch_output.path, "w", encoding="utf-8") as f:
+        if not os.path.exists(mmlu_branch_output_path):
+            os.makedirs(mmlu_branch_output_path)
+        mmlu_branch_output_file = (
+            Path(mmlu_branch_output_path) / "mmlu_branch_data.json"
+        )
+        with open(mmlu_branch_output_file, "w", encoding="utf-8") as f:
```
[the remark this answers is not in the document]
leseb: Yeah, no worries!

eval/mt_bench/components.py
MichaelClifford: Thank you, this is much better 😄 I will make this update in this PR as well.

eval/final/components.py
MichaelClifford: Yes, I think we originally pulled the metrics from the Artifacts, but then decided to use PVC's instead for consistency and speed. However, I think I replaced the Artifact steps with PVCs instead of adding it. But, I think it makes sense to maintain the eval steps' outputs as Artifacts too. Good catch. I will update the PR to make sure the final eval outputs are stored as artifacts. Thanks 😄

MichaelClifford: I've added `pvc_to_mt_bench_branch_op` and `pvc_to_mmlu_bench_op` to the `utils/components.py` and added them to the pipeline to store the final eval outputs as artifacts.

pipeline.py
```diff
@@ -463,6 +484,21 @@ def pipeline(
         model_pvc_delete_task = DeletePVC(pvc_name=model_pvc_task.output)
         model_pvc_delete_task.after(final_eval_task)
 
+        generate_metrics_report_task = generate_metrics_report_op()
+        generate_metrics_report_task.after(output_mt_bench_task, final_eval_task)
```
tumido, line 488: I don't think this requires to be run after `output_mt_bench_task`. That task only uploads artifacts. In fact both `output_mt_bench_task` and `generate_metrics_report_task` can run in parallel, they only read the same data.

tumido (commented):
I have one last suggestion, otherwise LGTM. 👍 🙌

pipeline.py
MichaelClifford: You are right. :) removing that requirement from `.after()`

tumido (changes requested):
It seems like you've forgot to include `pipeline.py` changes after rebase. As a result, the `pvc_to_mt_bench_branch_op` and `pvc_to_mmlu_branch_op` tasks are not used at all.

pipeline.py
```diff
@@ -432,6 +453,20 @@ def ilab_pipeline(
     model_pvc_delete_task = DeletePVC(pvc_name=model_pvc_task.output)
     model_pvc_delete_task.after(final_eval_task)
 
+    generate_metrics_report_task = generate_metrics_report_op()
+    generate_metrics_report_task.after(output_mt_bench_task, final_eval_task)
```
tumido, line 457: I think this depends on `final_eval_task` only, doesn't it?

pipeline.py
```diff
@@ -432,6 +453,20 @@ def ilab_pipeline(
...
+    generate_metrics_report_task.set_caching_options(False)
+    mount_pvc(
+        task=generate_metrics_report_task,
+        pvc_name=output_pvc_task.output,
+        mount_path="/output",
+    )
+
+    output_pvc_delete_task = DeletePVC(pvc_name=output_pvc_task.output)
+    output_pvc_delete_task.after(
+        output_model_task, output_mt_bench_task, final_eval_task
```
tumido, line 467: Can you please add the dependency on the new artifact upload tasks?

`output_mmlu_branch_task`, `output_mt_bench_branch_task`

I think it should also run only after on `generate_metrics_report_task` finishes, so maybe that's also a dependency...

tumido (changes requested):
Sorry to be annoying, just 2 last nits... otherwise LGTM 🙂 🙌

leseb (approved):

pipeline.py
MichaelClifford: yes it does 😄

pipeline.py
```diff
@@ -432,6 +453,22 @@ def ilab_pipeline(
...
+        task=generate_metrics_report_task,
+        pvc_name=output_pvc_task.output,
+        mount_path="/output",
+    )
+
+    output_pvc_delete_task = DeletePVC(pvc_name=output_pvc_task.output)
+    output_pvc_delete_task.after(
+        output_mmlu_branch_task,
+        output_mt_bench_branch_task,
+        generate_metrics_report_task,
```
tumido, line 469: I'm sorry to be so stubborn here. 😄  I still think this is not a correct list of dependencies. I think it should be:

```suggestion
        output_model_task,
        output_mt_bench_task,
        output_mmlu_branch_task,
        output_mt_bench_branch_task,
        generate_metrics_report_task,
```

tumido (changes requested):

pipeline.py
MichaelClifford: I appreciate your stubbornness on this PR 😃 Change made.

tumido (approved):
/lgtm
/approve

PR #242, opened 2024-12-18
set caching to False for output_mt_bench_task

MichaelClifford:
This PR sets `outpiut_mt_bench_task` caching option to `False`. Otherwise, this can lead to instances where the MT-bench data changes, but the MT-bench data artifacts do not get updated, causing a mismatch between the evaluation run and the evaluation values reported.


tumido (approved):
/lgtm
/approve

PR #243, opened 2024-12-20
Parameterize TLS configuration options

gmfrasca:
Update Pipeline to allow TLS Cert verification when connecting to Teacher- and Judge- models

## Description
- Add `_use_tls` switch params for sdg, mt_bench, and final_eval phases
- Source certs from teacher- and judge-server ConfigMaps (already used for other metadata)
- Update eval phases to create and use httpx client instead of a custom-patched openai client
  - (Required update after [update to ilab eval](https://github.com/instructlab/eval/commit/010ff8c11db4e2d3b4812ec8bdf6b6ef94442687#diff-e5e89be967cd71d6e6a091b6f72126552782a6da4abac9c0e0b9c9fac33b2745L124)
- Update README 

## How Has This Been Tested?
- Run pipeline with certs and _use_tls params enabled, verify successful connections to both teacher and judge models during SDG and Eval phases
- Run pipeline with certs turned off (and models requiring them), verify phases fail due to incorrect/missing certs error

## Merge criteria:


sdg/components.py
```diff
@@ -47,9 +48,10 @@ def sdg_op(
     model = getenv("model")
     endpoint = getenv("endpoint")
 
-    if sdg_ca_cert := getenv("SDG_CA_CERT_PATH"):
+    if use_tls:
```
MichaelClifford, line 51: Do we need to parametrize this? Instead can we do something like 
```suggestion
    if getenv("SDG_CA_CERT_PATH"):
```
This way we just default to TLS if there is a cert and not if there is not one? Would there be a case where the user would have a cert in their configmap, but would not want it to be used?

gmfrasca: So SDG_CA_CERT_PATH will always point to a truthy value (/tmp/cert/ca.crt), so this conditional will always trigger.  Wed have to actially check the path for existance of the file which is a bit more involved

MichaelClifford: Is it more involved then: 
```
sdg_ca_cert = getenv("SDG_CA_CERT_PATH")
if os.path.exists(sdg_ca_cert):
```

Or do we need to actually read the file and validate its contents too?

gmfrasca: i think we can do something like this, probably should ensure the file is actually populated / non-empty too, but that's an easy check

MichaelClifford (approved):
LGTM! Thanks @gmfrasca

PR #246, opened 2025-01-07
refactor: remove phase mocking and faked packages

### LINKED ISSUE #54: Re-Evaluate Mocked Phases
This is a somewhat long pipeline and some of the steps can take a significant amount of time. As a developer, I want to be able to work on and test different parts of the pipeline without waiting on all previous steps to complete first.  We currently have an approach of using "faked"  pipeline steps with the same names but different functionality. While developing, I've found that even with the faked steps moving data around can take a long time and hinder development speed.

I propose that we generate a long lasting PVC on our cluster that is pre-loaded with all of the interim datasets and models needed throughout the pipeline so that we can easily skip directly to whichever step is relevant for a developer without waiting on any previous data passing steps.

tumido:
Resolves #54

## Description
Since we've migrated to PVC data passing and did a lot of other optimizations, there's no use for the mocked phases anymore and nobody is using them. That makes this approach a tech debt which we should clean up.

In addition to that, I've reviewed PyLint exceptions and cleaned up some duplicate imports within individual components.


utils/components.py
```diff
@@ -1,5 +1,4 @@
 # type: ignore
-# pylint: disable=no-value-for-parameter,import-outside-toplevel,import-error,no-member,missing-function-docstring
```
leseb, line 2: Why is this gone?

training/components.py
```diff
@@ -1,5 +1,5 @@
 # type: ignore
-# pylint: disable=import-outside-toplevel,missing-function-docstring
```
leseb, line 2: Why is this being changed?

leseb (commented):
I think the removal of the mock/faked package and the relocation of various files for eval could have been handled as separate changes. Having both makes it harder to review but since it's there I don't want to block it, let's try to isolate the changes into different PRs next time :). Thanks!

I just left two small questions otherwise LGTM. Thanks!

utils/components.py
tumido: It seems it was not needed anymore... Or at least my local machine does not report any errors anymore. 🤷

training/components.py
tumido: Because we need `import-error` in here due to:
image


And it seems like `missing-function-docstring` is just info severity + we do not declare it anywhere else anyways (except for the one mention I've removed below)...

leseb (approved):

PR #248, opened 2025-01-07
Remove hardcoded training image refernce

gmfrasca:
- Already defined as a const so just use that for maintainability

## Merge criteria:


gmfrasca:
actually, since this is encapsulated in a component we'll need to plumb the value through, which is a bit more complicated.  Still should do this, otherwise we have hardcoded image refs

tumido:
Yup, good idea 🙂 Is this still WIP?

For the context we've been trying to do the same (and other stuff to refactor the KFTO integration) via #231

There's a thread to this particular issue, but I think the solution is the same as yours 🙂 

https://github.com/opendatahub-io/ilab-on-ocp/pull/231#discussion_r1887206093

(BTW I think your PR should be merged anyways, since @Shreyanand is still blocked on the notebook image not being available I think.)

PR #251, opened 2025-01-14
WIP: Pass RHELAI_IMAGE to pytorchjobs via envvar

gmfrasca:
Pass the pytorchjob base image reference via environment variable

## Description
Get around hardcoded RHELAI Image used for Training PyTorchJobs by passing the const (which is already defined in `util/consts.py`) to the manifest op via an env var.  This will mean that *all* image references are now centralized in `util/consts.py`

## Merge criteria:

### DIFF
--- a/pipeline.py
+++ b/pipeline.py
@@ -36,6 +36,7 @@ from utils import (
     pvc_to_model_op,
     pvc_to_mt_bench_op,
 )
+from utils.consts import RHELAI_IMAGE  # To pass to training PyTorchJobs
 
 TEACHER_CONFIG_MAP = "teacher-server"
 TEACHER_SECRET = "teacher-server"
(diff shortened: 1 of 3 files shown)


gmfrasca:
This would supersede https://github.com/opendatahub-io/ilab-on-ocp/pull/248 .  

I think this is a bit cleaner, albeit less direct, way to get the RHELAI_IMAGE ref to the manifest op rather than plumbing the value directly through via component and pipeline params

leseb (commented):
I believe https://github.com/opendatahub-io/ilab-on-ocp/pull/248 was more straightforward. Using an environment variable feels more like a workaround.

From my perspective, components should be self-contained and “portable.” Introducing this environment variable as a dependency creates a hidden requirement that isn’t explicitly part of the function definition. If we ever need to move this component to another pipeline, we’ll have to set the environment variable again. In contrast, having all the required values defined within the function itself makes the dependency much clearer.

PR #257, opened 2025-01-23
fix: sync generated files

tumido:
This PR is a result of simply running `make` on the `main` branch.

It seems like the missing changeset comes from https://github.com/opendatahub-io/ilab-on-ocp/pull/243

Ignore CI failure, it's false positive (it complains that I'm modifying generated files only and nothing has changed in the actual sources, which is what we want in this particular case)

### DIFF
--- a/standalone/standalone.py
+++ b/standalone/standalone.py
@@ -1132,20 +1132,27 @@ def sdg_op(
     sdg_path: str = "/data/sdg",
     sdg_sampling_size: float = 1.0,
 ):
-    from os import getenv, path
+    import os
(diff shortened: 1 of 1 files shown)


leseb (approved):
Unfortunately, this [condition](https://github.com/opendatahub-io/ilab-on-ocp/blob/main/.github/workflows/pre_commit.yaml#L46) is not working since it's based on the PR creation, when the PR event was triggered. Anyway, I have the same result locally. LGTM!

HumairAK:
cc @gmfrasca

PR #258, opened 2025-01-27
Add dockerfile for Python Runtime Image & parameterize images in makefile via env

HumairAK:
## Description

This change does the following: 

* Add dockerfile to be used as the python_image for runtime. The image includes the kfp/kubeflow training dependencies 
  * The image needs to have all the dependencies required to make the pipeline(s)
* Include requirements.txt to leverage for system wide python package installs
* Add requirements.txt check to ci and readme 
* Parameterize image files at makefile level 

The following change will allow us to build the pipeline during a runtime init container step, that will allow us to consume updated images in an automated way in the future. 

Image defaults in makefile are kept as SHA for now, once automation is enabled, we should set these to floating tags. 

Instead of adding `requirements.txt` another option is to just leverage `uv`, this would require us to include this as an additional package dependency in the image, and I _think_ do a `uv pip install --system`, I'm not that familiar with uv however. 

Jira: https://issues.redhat.com/browse/RHOAIENG-18413

## How Has This Been Tested?
Ran the pipeline with the built image set to PYTHON Image.

## Merge criteria:


mprahl (approved):

HumairAK:
fyi @MichaelClifford / @leseb / @tumido / @Shreyanand / et al

Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM registry.redhat.io/ubi8/python-312:1
```
mprahl, line 1: A small nitpick, but if you use `registry.access.redhat.com` instead of `registry.redhat.io` for UBI images, you don't need authentication to pull it down. It just simplifies things.

HumairAK: you're right, this was updated here: https://github.com/opendatahub-io/ilab-on-ocp/pull/262/files

PR #260, opened 2025-01-28
Use TLS when Cloning Taxonomy Tree if CACert provided

gmfrasca:
TLS-enabled git operations for Taxonomy

## Description
If a CA Cert is provided in the form of a 'taxonomy-ca.crt' in the `teacher-server` ConfigMap, use TLS and said CA Cert to clone and fetch the taxonomy tree.

## How Has This Been Tested?
To test, using a git repo that is accessed using a self-signed CA cert:
1. Import pipeline.yaml
2. For now, ensure `teacher-server` ConfigMap does *NOT* have a key named `taxonomy-ca.crt`
3. Run pipeline with default taxonomy repo pointer [Expected Result: Git Clone task should succeed]
4. Run with the taxonomy_repo parameter pointed to a git repo that can only be accessed using a self-signed CA cert.  [Expected Result: Fails to clone]
5. Add the CA Certificate to `teacher-server` ConfigMap under the key `taxonomy-ca.crt`
6. Repeat Step 3. [Expected Result: Git Clone task should succeed]
7. Repeat Step 4. [Expected Result: Git Clone task should succeed]

## Merge criteria:

### DIFF
--- a/sdg/components.py
+++ b/sdg/components.py
@@ -13,27 +13,38 @@ def git_clone_op(
     repo_pr: Optional[int],
     repo_url: Optional[str],
     taxonomy_path: str = "/data/taxonomy",
-    ca_cert_path: Optional[str] = None,
 ):
-    import os
-
-    additional_clone_params = ""
-    additional_config_cmds = ""
-
-    if ca_cert_path and os.path.exists(f"{ca_cert_path}") and (os.path.getsize(f"{ca_cert_path}") > 0):
-        full_ca_path = os.path.abspath(f"{ca_cert_path}")
-        additional_clone_params = f"-c http.sslVerify=true -c http.sslCAInfo={full_ca_path}"
-        additional_config_cmds = f"git config http.sslVerify true && git config http.sslCAInfo {full_ca_path} &&"
-
     return dsl.ContainerSpec(
         TOOLBOX_IMAGE,
         ["/bin/sh", "-c"],
         [
-            f"git clone {additional_clone_params} {repo_url} {taxonomy_path} && cd {taxonomy_path} && {additional_config_cmds}"
-            + f'if [ -n "{repo_branch}" ]; then '
-            + f"git fetch origin {repo_branch} && git checkout {repo_branch}; "
-            + f'elif [ -n "{repo_pr}" ] && [ {repo_pr} -gt 0 ]; then '
-            + f"git fetch origin pull/{repo_pr}/head:{repo_pr} && git checkout {repo_pr}; fi "
+            f"""
+            # Increase logging verbosity
+            set -x &&
+
+            # Add TLS Parameters if CA Cert exists and is non-zero size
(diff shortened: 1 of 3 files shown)


jgarciao:
As the [recommended way to add CA bundles in RHOAI](https://docs.redhat.com/en/documentation/red_hat_openshift_ai_self-managed/2.16/html/installing_and_uninstalling_openshift_ai_self-managed/working-with-certificates_certs) is to add them in the DSCInitialization (in .spec.trustedCABundle.customCABundle) and those bundles are already mounted in the pod running the task, I think the ilab pipeline should follow the same pattern and not use a different configmap to store the certificates

sdg/components.py
```diff
@@ -14,11 +14,32 @@ def git_clone_op(
...
+    # Add TLS Parameters if CA Cert exists
+    taxonomy_ca_cert_path = os.getenv("TAXONOMY_CA_CERT_PATH")
+    use_tls = (
+        taxonomy_ca_cert_path
+        and os.path.exists(taxonomy_ca_cert_path)
+        and (os.path.getsize(taxonomy_ca_cert_path) > 0)
+    )
+
+    if use_tls:
+        full_ca_path = os.path.abspath(f"{taxonomy_ca_cert_path}")
```
mprahl, line 32: Optional:
```suggestion
        full_ca_path = os.path.abspath(taxonomy_ca_cert_path)
```

mprahl (approved):

sdg/components.py
gmfrasca: That variable is actually not a string - i believe its an InputParameterChannel or some other Pipelines construct, which abspath doesnt accept as a string-like type.  Because of that, we need to expand it, as casting this specific type if it is None actually expands to the string "None" which would incorrectly pass this check

mprahl (approved):

sdg/components.py
```diff
@@ -18,11 +18,33 @@ def git_clone_op(
...
+
+            # Clone Taxonomy Repo
+            git clone $ADDITIONAL_CLONE_PARAMS {repo_url} {taxonomy_path} &&
+            cd {taxonomy_path} &&
+
+            # Run additional configuration if TLS certs provided
+            if [ -s "$TAXONOMY_CA_CERT_PATH" ]; then
+                git config http.sslVerify true &&
+                git config http.sslCAInfo $TAXONOMY_CA_CERT_PATH
+            fi &&
```
HumairAK, line 39: can't we just do this in the beginning and avoid the need for `ADDITIONAL_CLONE_PARAMS` ?

gmfrasca: no, this is repo-specific so need to do the `git clone` first

gmfrasca: (and doing a global config via `--global` hits permission issues because this is run in a container)

HumairAK:
cc @tumido / @Shreyanand / @leseb / @MichaelClifford  / et al

PR #267, opened 2025-02-07
Add verification step to ilab pipeline

rimolive:
As part of [RHOAIENG-18882](https://issues.redhat.com/browse/RHOAIENG-18882), this PR implements a new component in ilab pipeline to check judge and teacher model availability

## Description
This PR adds a new component in ilab pipeline to check the following information:
* Teacher model validation
* Judge model validation
* Model Registry connectivity
* OCI Model car parameters validation
* Kubeflow Training Operator installation
* Taxonomy git repo validation

## How Has This Been Tested?
**NOTE:** The ilab pipeline yaml file was compiled using https://github.com/kubeflow/pipelines/pull/11621. Note that the test involves update Driver image with the changes in this PR

* Upload `pipeline.yaml`
* Create a run with the default parameters
* A sub DAG will be present with 6 steps:
  * Teacher model validation
  * Judge model validation
  * Model Registry connectivity
  * OCI Model car parameters validation
  * Kubeflow Training Operator installation
  * Taxonomy git repo validation
* Create a run:
  * **NOTE:** This step requires that both judge and teacher models are up and running, model Registry is installed and configured properly, KFTO is installed, and there's a valid dockerconfigjson or dockercfg secret

## Merge criteria:

- [X] The commits are squashed in a cohesive manner and have meaningful messages.
- [X] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [X] The developer has manually tested the changes and verified that the changes work


mprahl:
@rimolive could you please rebase the PR?

utils/components.py
```diff
@@ -61,3 +69,311 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+    import requests
+    from kubernetes import client
+    from kubernetes.client.rest import ApiException
+
+    sa_path = "/var/run/secrets/kubernetes.io/serviceaccount"
+    configuration = client.Configuration()
+    configuration.api_key["authorization"] = open(f"{sa_path}/token").readline()
+    configuration.api_key_prefix["authorization"] = "Bearer"
+    configuration.host = f"https://{os.environ['KUBERNETES_SERVICE_HOST']}"
+    configuration.ssl_ca_cert = f"{sa_path}/ca.crt"
```
mprahl, line 90:
```suggestion
    import kubernetes.config

    kubernetes.config.load_incluster_config()
```

utils/components.py
```diff
@@ -61,3 +69,311 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+
+    with client.ApiClient(configuration) as api_client:
+        core_api = client.CoreV1Api(api_client)
+
+        try:
+            config_map = core_api.read_namespaced_config_map(cm_name, namespace)
+            print(f"Reading configmap {cm_name} data...")
+            model_endpoint = config_map.data["endpoint"]
+            model_name = config_map.data["model"]
+        except ApiException as e:
```
mprahl, line 105: Could you please also catch `KeyError` exceptions in case the ConfigMap and Secret exist but the keys aren't specified?

utils/components.py
```diff
@@ -61,3 +69,311 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+            ############################################################################################################
+            """)
+            sys.exit(1)
+
+    request_auth = {"Authorization": f"Bearer {model_api_key}"}
+    request_body = {
+        "model": model_name,
+        "messages": [{"role": "user", "content": "tell me a funny joke."}],
+    }
+    resp = requests.post(
```
mprahl, line 130: Optional: It'd be nice to at least try this a few times before failing in case of a network flake.

utils/components.py
```diff
@@ -61,3 +69,311 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+    packages_to_install=["model-registry==0.2.13", "kserve==0.13"],
+)
+def test_model_registry(model_registry_endpoint: str):
+    from model_registry import ModelRegistry
+
+    registry = ModelRegistry(
+        server_address=model_registry_endpoint,
+        author="ilab pipeline",
+        is_secure=False,
+        user_token=open(
```
mprahl, line 162: Optional: It'd be nice to open the file in a context manager (i.e. `with open`) but it doesn't particularly matter here since the function ends shortly after.

utils/components.py
```diff
@@ -61,3 +69,311 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+@dsl.component(base_image=PYTHON_IMAGE)
+def test_training_operator():
+    import os
+    import sys
+
+    from kubernetes import client
+    from kubernetes.client.rest import ApiException
+
+    sa_path = "/var/run/secrets/kubernetes.io/serviceaccount"
+    configuration = client.Configuration()
```
mprahl, line 184: Same comment as above. Let's use `kubernetes.config.load_incluster_config()`.

utils/components.py
```diff
@@ -61,3 +69,311 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+    configuration.api_key["authorization"] = open(f"{sa_path}/token").readline()
+    configuration.api_key_prefix["authorization"] = "Bearer"
+    configuration.host = f"https://{os.environ['KUBERNETES_SERVICE_HOST']}"
+    configuration.ssl_ca_cert = f"{sa_path}/ca.crt"
+    namespace = open(f"{sa_path}/namespace").readline()
+
+    with client.ApiClient(configuration) as api_client:
+        api_instance = client.CustomObjectsApi(api_client)
```
mprahl, line 192: Could you directly check if the CRD object is available rather check if there is a CR?

utils/components.py
```diff
@@ -61,3 +69,311 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+    import base64
+    import json
+    import os
+    import sys
+
+    from kubernetes import client
+    from kubernetes.client.rest import ApiException
+
+    sa_path = "/var/run/secrets/kubernetes.io/serviceaccount"
+    configuration = client.Configuration()
```
mprahl, line 226: Same comment as above. Let's use `kubernetes.config.load_incluster_config()`.

utils/components.py
```diff
@@ -61,3 +69,311 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+    configuration.api_key_prefix["authorization"] = "Bearer"
+    configuration.host = f"https://{os.environ['KUBERNETES_SERVICE_HOST']}"
+    configuration.ssl_ca_cert = f"{sa_path}/ca.crt"
+    namespace = open(f"{sa_path}/namespace").readline()
+
+    if not sdg_base_model.startswith("oci://"):
+        # If SDG Model Base is not an OCI image, just inform user and quit
+        print("""
+        #################### INFO ###########################
+        # Model is not OCI-compliant. Skipping this step... #
```
mprahl, line 237:
```suggestion
        # Model {sdg_base_model} is not a Modelcar. Skipping this step... #
```

utils/components.py
```diff
@@ -61,3 +69,311 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+    teacher_cm_name: str,
+    teacher_secret_name: str,
+    model_registry_endpoint: str,
+):
+    """
+    Pre-validation checks for the InstructLab pipeline.
+    """
+    import os
+
+    # TODO: Is it required to disable caching for all validation steps?
```
mprahl, line 328: I think the answer is yes. Did you find something different?

utils/components.py
```diff
@@ -61,3 +69,311 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+            fi
+
+            # ls-remote will fail if repo is not valid
+            git ls-remote {sdg_repo_url} > /dev/null;
+            """
+        ],
+    )
+
+
+@dsl.pipeline(display_name="Pre-requisite check")
```
mprahl, line 312:
```suggestion
@dsl.pipeline(display_name="Prerequisites check")
```

utils/components.py
```diff
@@ -61,3 +69,311 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+def pre_requisites_check_op(
```
mprahl, line 313:
```suggestion
def prerequisites_check_op(
```

utils/components.py
```diff
@@ -61,3 +69,311 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+        # Model Server {model_name} is up and running. #
+        ################################################
+        """)
+
+
+@dsl.component(
+    base_image=PYTHON_IMAGE,
+    packages_to_install=["model-registry==0.2.13", "kserve==0.13"],
+)
+def test_model_registry(model_registry_endpoint: str):
```
mprahl, line 155: We will need to make this optional if `output_model_registry_name` (a pipeline input in the main branch now) is not specified.

utils/components.py
rimolive: No, we should disable caching as the subsequent run with same parameters will skip validation. Removing this TODO.

utils/components.py
rimolive: I can't find a better way to check directly if CRD is present. If you have any ideas how to do that, I can make the change.

mprahl: Sorry, what I meant was the equivalent of `kubectl get crd pytorchjobs.kubeflow.org`.

utils/components.py
```diff
@@ -298,3 +301,302 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+):
+    """
+    Pre-validation checks for the InstructLab pipeline.
+    """
+    import os
+
+    ## Validate judge information
+    test_judge_model_op = test_judge_teacher_models(secret_name=eval_judge_secret)
+    use_config_map_as_volume(
+        test_judge_model_op, "kube-root-ca.crt", mount_path="/tmp/cert"
```
mprahl, line 573: This should use the `JUDGE_CONFIG_MAP` rather than hardcoding `kube-root-ca.crt` like how we do for `run_mt_bench_op`.

utils/components.py
```diff
@@ -298,3 +301,302 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
             f"ilab --config=DEFAULT model download --repository {repository} --release {release} --model-dir {base_model.path}"
         ],
     )
+
+
+@dsl.component(base_image=RUNTIME_GENERIC_IMAGE)
+def test_judge_teacher_models(secret_name: str):
```
mprahl, line 307: I'd prefer this name or similar because the current name sounds like it checks both in one execution of the function.
```suggestion
def test_model_connection(secret_name: str):
```

utils/components.py
```diff
@@ -298,3 +301,302 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+        if resp.status_code != 200:
+            print(f"Model Server {model_name} is not available. Attempt {i}/3...")
+            time.sleep(30)
+        else:
+            print(f"""
+            ################### INFO #######################
+            # Model Server {model_name} is up and running. #
+            ################################################
+            """)
+            pass
```
mprahl, line 367: Did you mean this instead?
```suggestion
            break
```

utils/components.py
```diff
@@ -298,3 +301,302 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+
+
+@dsl.component(base_image=RUNTIME_GENERIC_IMAGE)
+def test_model_registry(
+    model_registry_endpoint: Optional[str],
+    model_name: Optional[str],
+    model_version: Optional[str],
+):
+    import sys
+
```
mprahl, line 383: Could you return early if `model_registry_endpoint` is not defined since registering to the Model Registry is optional?

utils/components.py
```diff
@@ -298,3 +301,302 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+    from model_registry import ModelRegistry
+
+    try:
+        model_registry = ModelRegistry(
+            server_address=model_registry_endpoint,
+            author="ilab pipeline",
```
mprahl, line 389: Not that I think it matters much
```suggestion
            author="InstructLab Pipeline",
```

utils/components.py
```diff
@@ -298,3 +301,302 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+            is_secure=False,
```
mprahl, line 390: We want this to fail now if TLS verification fails because that means the trusted platform CA bundle does not have the CA that signed the Model Registry.
```suggestion
```

utils/components.py
```diff
@@ -298,3 +301,302 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+            user_token=open(
+                "/var/run/secrets/kubernetes.io/serviceaccount/token"
+            ).readline(),
+        )
+        if model_registry.get_model_version(model_name, model_version) is not None:
+            print(f"""
+            ####################################### ERROR #######################################
+            # There is already a model version registered. You cannot overwrite a model version #
```
mprahl, line 398:
```suggestion
            # The version {model_version} for model {model_name} is already. You cannot overwrite a model version. #
```

utils/components.py
```diff
@@ -298,3 +301,302 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+    try:
+        model_registry = ModelRegistry(
+            server_address=model_registry_endpoint,
+            author="ilab pipeline",
+            is_secure=False,
+            user_token=open(
+                "/var/run/secrets/kubernetes.io/serviceaccount/token"
+            ).readline(),
+        )
+        if model_registry.get_model_version(model_name, model_version) is not None:
```
mprahl, line 395: You need to also guard against it raising a `StoreError` exception because if the model doesn't exist, that's also okay since the first version being registered will register the model as well.

utils/components.py
```diff
@@ -298,3 +301,302 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+    config.load_incluster_config()
+
+    if output_oci_model_uri is None:
+        print(f"""
+        ############################## INFO ##################################
+        # Parameter output_oci_model_uri not provided. Skipping this step... #
+        ######################################################################
+        """)
+        sys.exit(0)
+    elif not output_oci_model_uri.startswith("oci://"):
```
mprahl, line 475: This is actually okay because `upload_model_op` adds the `oci://` prefix if it's missing. This is okay because the pipeline parameter includes `oci` in the name, so it implies that any user value for it means they want to push the model as a Modelcar.

utils/components.py
```diff
@@ -298,3 +301,302 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+        core_api = client.CoreV1Api(api_client)
+        try:
+            secret = core_api.read_namespaced_secret(sdg_oci_docker_secret, namespace)
+            print(f"Reading secret {sdg_oci_docker_secret} data...")
+            if secret.type == "kubernetes.io/dockerconfigjson":
+                # handle authentication if secret provided is kubernetes.io/dockerconfigjson
+                docker_config_json = json.loads(
+                    base64.b64decode(secret.data[".dockerconfigjson"]).decode("utf-8")
+                )
+                if registry_name not in docker_config_json["auths"]:
```
mprahl, line 497: This doesn't seem to be permissive enough as wildcards and slashes are accepted:
https://kubernetes.io/docs/concepts/containers/images/#config-json

I found the Kubernetes code for this where `glob` comes from Docker config file and target is the image:
https://github.com/kubernetes/kubernetes/blob/4468565250c940bbf70c2bad07f2aad387454be1/pkg/credentialprovider/keyring.go#L183-L233

This might be too much to reimplement using Python and if it is, I'd be okay with just verifying the secret exists and has a valid Docker config.

utils/components.py
```diff
@@ -298,3 +301,302 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+            set -x &&
+
+            # Add TLS Parameters if CA Cert exists and is non-zero size
+            ADDITIONAL_CLONE_PARAMS=""
+            if [ -s "$TAXONOMY_CA_CERT_PATH" ]; then
+                ADDITIONAL_CLONE_PARAMS="-c http.sslVerify=true -c http.sslCAInfo=$TAXONOMY_CA_CERT_PATH"
+            fi
+
+            # ls-remote will fail if repo is not valid
+            git ls-remote {sdg_repo_url} > /dev/null;
```
mprahl, line 548: Optional: It'd be nice to retry this on failure since you do it elsewhere in the code, but not required since it's easy enough to retry due to a network flake.

mprahl (changes requested):
I left some comments. Let me know if I can help!

pipeline.py
```diff
@@ -187,6 +199,25 @@ def ilab_pipeline(
         size=k8s_storage_size,
         storage_class_name=k8s_storage_class_name,
     )
+    sdg_input_pvc_task.after(prerequisites_check_task)
+    git_clone_task = git_clone_op(
```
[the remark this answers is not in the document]
rimolive: This is the actual git clone task for taxonomy. I did not touch in that code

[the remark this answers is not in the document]
rimolive: This is the actual git clone task for taxonomy. I did not touch in that code

utils/components.py
```diff
@@ -298,3 +301,362 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+    model_registry_endpoint: Optional[str],
+    model_name: Optional[str],
+    model_version: Optional[str],
+):
+    import sys
+
+    from model_registry import ModelRegistry
+    from model_registry.exceptions import StoreError
+
+    if model_registry_endpoint is None:
```
mprahl, line 387: That way we can skip empty strings as well.
```suggestion
    if not model_registry_endpoint:
```

pipeline.py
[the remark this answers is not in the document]
rimolive: From what I can see, the `git_clone_task` task just clone the taxonomy repo to a PVC, and it gets reused by the other tasks. This is not for a prerequisite check but to prepare the environmnent to run the ilab steps.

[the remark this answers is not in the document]
rimolive: Ah, I see it now. This task was removed. I'm going to remove it as well.

utils/components.py
```diff
@@ -298,3 +307,359 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+        try:
+            secret = core_api.read_namespaced_secret(secret_name, namespace)
+            print(f"Reading secret {secret_name} data...")
+            model_api_key = base64.b64decode(secret.data["api_key"]).decode("utf-8")
+            model_name = base64.b64decode(secret.data["model_name"]).decode("utf-8")
+            model_endpoint = base64.b64decode(secret.data["endpoint"]).decode("utf-8")
+        except (ApiException, KeyError) as e:
+            print(f"""
+            ####################################################### ERROR ################################################################
+            # Error reading {secret_name}. Ensure you created a secret with this name in namespace {namespace} and has 'api_key' present #
```
mprahl, line 346: The error message is a bit misleading because it also checks for `model_name` and `endpoint` and the error message doesn't mention those.

utils/components.py
```diff
@@ -298,3 +307,359 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+
+    from kubernetes import client, config
+    from kubernetes.client.rest import ApiException
+
+    def url_match(sources, target):
+        import re
+        import urllib
+
+        for source in sources:
+            source_split = source.split(".")
```
mprahl, line 507: It looks like Kubernetes also checks that the port matches, so we should too.

utils/components.py
```diff
@@ -298,3 +307,359 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+            target_split = target.split(".")
+            if len(source_split) != len(target_split):
+                return False
+            for idx, val in enumerate(source_split):
+                regex = source_split[idx].replace("*", ".*")
```
mprahl, line 512: I don't think this will work because `source_split[idx]` may have unescaped RegEx characters. You could use `regex.escape` but this might be a bit error prone since Kubernetes uses [filepath.Match](https://pkg.go.dev/path/filepath#Match).

pipeline.py
```diff
@@ -240,13 +252,15 @@ def ilab_pipeline(
     model_source_task = dsl.importer(
         artifact_uri=sdg_base_model, artifact_class=dsl.Model
     )
+    model_source_task.after(prerequisites_check_task)
```
mprahl, line 255: After the rebase, you'll need to add `model_tokenizer_source_task.after(prerequisites_check_task)` as well.

utils/components.py
```diff
@@ -298,3 +307,338 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+        )
+        if registry.get_model_version(model_name, model_version) is not None:
+            print(f"""
+            ###################################################### ERROR ######################################################
+            # The version {model_version} for model {model_name} is already registered. You cannot overwrite a model version. #
+            ###################################################################################################################
+            """)
+            sys.exit(1)
+    except StoreError as store:
+        # Do nothing, just to avoid this exception to return failure
```
mprahl, line 445:
```suggestion
        # The model has no versions registered.
        # Do nothing, just to avoid this exception to return failure
```

utils/components.py
mprahl: I just realized that the pipeline-runner service account won't have access to query the CRD, so let's just use `list_namespaced_custom_object` with the pod's current namespace.

utils/components.py
```diff
@@ -298,3 +307,339 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+            secret = core_api.read_namespaced_secret(sdg_oci_docker_secret, namespace)
+            print(f"Reading secret {sdg_oci_docker_secret} data...")
+            if secret.type == "kubernetes.io/dockerconfigjson":
+                # handle authentication if secret provided is kubernetes.io/dockerconfigjson
+                docker_config_json = json.loads(
+                    base64.b64decode(secret.data[".dockerconfigjson"]).decode("utf-8")
+                )
+                print(f"""
+                ######################## INFO ###########################
+                # OCI Secret has auth token present for {registry_name} #
```
mprahl, line 533: Since we aren't checking the registry name anymore, let's remove the registry name from the message since we didn't actually verify it.

utils/components.py
```diff
@@ -298,3 +307,339 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+                #########################################################
+                """)
+            elif secret.type == "kubernetes.io/dockercfg":
+                # handle authentication if secret provided is kubernetes.io/dockercfg
+                dockercfg_json = json.loads(
+                    base64.b64decode(secret.data[".dockercfg"]).decode("utf-8")
+                )
+                print(f"""
+                ######################## INFO ###########################
+                # OCI Secret has auth token present for {registry_name} #
```
mprahl, line 543: Since we aren't checking the registry name anymore, let's remove the registry name from the message since we didn't actually verify it.

utils/components.py
```diff
@@ -298,3 +307,340 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+        RUNTIME_GENERIC_IMAGE,
+        ["/bin/sh", "-c"],
+        [
+            f"""
+            # Increase logging verbosity
+            set -x &&
+
+            # Add TLS Parameters if CA Cert exists and is non-zero size
+            ADDITIONAL_CLONE_PARAMS=""
+            if [ -s "$TAXONOMY_CA_CERT_PATH" ]; then
```
mprahl, line 575: This is fine for now but will need to change based on #269

utils/components.py
```diff
@@ -298,3 +307,340 @@ def ilab_importer_op(repository: str, release: str, base_model: dsl.Output[dsl.M
...
+
+    with client.ApiClient() as api_client:
+        api_instance = client.CustomObjectsApi(api_client)
+
+        group = "kubeflow.org"
+        version = "v1"
+        plural = "pytorchjobs"
+
+        try:
+            api_response = api_instance.list_cluster_custom_object(
```
mprahl, line 479: As discussed in Slack, this needs to be `list_namespaced_custom_object` since PyTorchJobs are namespace scoped.

pipeline.py
```diff
@@ -240,13 +252,16 @@ def ilab_pipeline(
...
+    model_tokenizer_source_task.after(prerequisites_check_task)
```
mprahl, line 256: Optional: It'd be nice if this line was directly below `model_tokenizer_source_task = dsl.importer(` since it makes the code easier to read.

mprahl (approved):

PR #270, opened 2025-02-18
Install olot, skopeo, and model-registry in the generic runtime image

mprahl:
## Description

This will be required for building the Modelcar image in a future PR.

## Merge criteria:


pyproject.toml
```diff
@@ -10,6 +10,8 @@ dependencies = [
     "kfp-kubernetes>=1.4.0",
     "kfp==2.9.0",
     "kubernetes==30.1.0",
+    "model-registry==0.2.14",
```
rimolive, line 13: According to [Model Registry documentation](https://www.kubeflow.org/docs/components/model-registry/getting-started/#setup), you will need to include kserve too

Dockerfile
```diff
@@ -9,6 +9,8 @@ COPY ${SOURCE_CODE} .
 
 USER root
 
+RUN dnf install -y skopeo && dnf clean all
```
rimolive, line 12: Can we merge this line with line 14? It will be one less layer in the container image.

rimolive (commented):
Left some comments. Thanks @mprahl

dandawg (commented):
/lgtm

pyproject.toml
mprahl: It's not required unless you are creating an inference endpoint from the registered model. My testing of registering a model doesn't require it.

PR #271, opened 2025-02-25
add new input parameters

HumairAK:
## Description of changes 
Add inputs for v4 api: [[1]](https://docs.google.com/spreadsheets/d/16G5r6tV1TsgKbk-WoRoe0ZW9LsCWp093rnVDfySM33s/edit?gid=1780396156#gid=1780396156)

the following inputs were added:
* sdg/eval secrets
* model registration inputs
* input/output model info
* hardware inputs for training/eval
* renaming node/nproc variables to be easier to understand

these inputs don't do anything yet, functionality
soon to follow.

## Merge criteria:

### DIFF
--- a/pipeline.py
+++ b/pipeline.py
@@ -70,20 +70,36 @@ JUDGE_CA_CERT_PATH = "/tmp/cert"
     description="InstructLab pipeline",
 )
 def ilab_pipeline(
+    # Model I/O
+    input_model_uri: str,
+    output_oci_model_uri: str = "",
+    output_oci_registry_secret: str = None,
+    output_model_name: str = None,
+    output_model_version_name: str = None,
+    output_model_registry_name: str = None,
+    output_model_registry_namespace: str = "rhoai-model-registries",
+    output_modelcar_base_image: str = "registry.access.redhat.com/ubi9-micro:latest",
     # SDG phase
-    sdg_repo_url: str = "https://github.com/instructlab/taxonomy.git",
+    sdg_repo_url: str = None,
+    sdg_repo_secret: str = "taxonomy-repo-secret",
     sdg_repo_branch: Optional[str] = None,
     sdg_repo_pr: Optional[
         int
     ] = None,  # FIXME: https://issues.redhat.com/browse/RHOAIRFE-467
-    sdg_base_model: str = "s3://<BUCKET>/<PATH_TO_MODEL>",
+    sdg_teacher_secret: str = "teacher-secret",
+    sdg_base_model: str = None,
     sdg_scale_factor: int = 30,  # https://github.com/instructlab/instructlab/blob/v0.21.2/tests/testdata/default_config.yaml#L125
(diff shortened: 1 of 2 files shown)


pipeline.py
```diff
@@ -70,20 +70,36 @@
     description="InstructLab pipeline",
 )
 def ilab_pipeline(
+    # Model I/O
+    input_model_uri: str,
+    output_oci_model_uri: str,
+    output_oci_registry_secret: str,
+    output_model_name: str,
+    output_model_version: str,
+    output_model_registry_name: str,
```
mprahl, line 79: I think this should make them all optional as we agreed on since the user can just let the trained model be uploaded to S3.
```suggestion
    output_oci_model_uri: str = None,
    output_oci_registry_secret: str = None,
    output_model_name: str = None,
    output_model_version: str = None,
    output_model_registry_name: str = None,
```

pipeline.py
```diff
@@ -70,20 +70,36 @@
     description="InstructLab pipeline",
 )
 def ilab_pipeline(
+    # Model I/O
+    input_model_uri: str,
+    output_oci_model_uri: str,
+    output_oci_registry_secret: str,
+    output_model_name: str,
+    output_model_version: str,
```
mprahl, line 78: Optional: I think UX had made the comment that version name is clearer than just version so something like `output_model_version_name` may be clearer.

Gkrumbach07:
these look good to me.

pipeline.py
```diff
@@ -70,20 +70,36 @@
     description="InstructLab pipeline",
 )
 def ilab_pipeline(
+    # Model I/O
+    input_model_uri: str,
+    output_oci_model_uri: str = None,
```
mprahl, line 75: A slight implementation detail here is that `dsl.If` doesn't support `None` as a check, so we'll likely need to do something like this:
```suggestion
    output_oci_model_uri: str = "",
```

pipeline.py
```diff
@@ -103,23 +119,42 @@ def ilab_pipeline(
     final_eval_few_shots: int = 5,  # https://github.com/instructlab/instructlab/blob/v0.21.2/tests/testdata/default_config.yaml#L56
     final_eval_batch_size: str = "auto",  # https://github.com/instructlab/instructlab/blob/v0.21.2/tests/testdata/default_config.yaml#L52
     final_eval_merge_system_user_message: bool = False,  # https://github.com/instructlab/instructlab/blob/v0.21.2/src/instructlab/model/evaluate.py#L474
+    # General Evaluation Inputs
```
gmfrasca, line 122: i wonder if we should change the other eval params to conform to the prefix here, ie `eval_final_xyz`, `eval_mt_bench_xyz`, etc

HumairAK: yeah I wasn't sure either, let's leave that for ux to figure out, they have a column for their suggestion in the api doc now

mprahl (approved):

PR #272, opened 2025-02-25
add hardware parameters, secret parameters, and taxonomy repo authentication

HumairAK:
## Description
Parameterized the secret names for sdg/judge. These have some duplicated code in the form of a copied function `fetch_secret`, some notes: 

* The code is duplicated because we do not want to have a common component that does this, otherwise we would be storing user token info in persistent object storage
* We do not use the kubernetes python package because this is not available in rhelai image, and it doesn't make sense to request to add it there since we'll replace this code later anyway. Instead we opt for rest calls to the k8s api server. 
* We expose the underlying pod specs for training because the higher level abstraction `get_pod_template_spec` does not expose underlying hardware fields 
* I don't think we need `ISTIO_SIDECAR_INJECTION` in the training pod but I've kept it to be consistent with the underlying `get_pod_template_spec` implementation. 

The PR also adds parameterizes the secret for taxonomy repo, this adds auth management for this repo, user can provide ssh or username/pat authentication. This has various edge cases for consideration, and I've tested the following: 

- [x] Self-Hosted (https) Self-Signed 
	- [x] Public
	- [x] Private
- [ ]  Github.com (well-known CA)
	- [x] https
		- [x] public (instructlab/taxonomy.git)
		- [ ] private (personal private fork)
			- [ ] this would require github PAT + username
	- [x] ssh
		- [x] public (non-anonymous, with ssh secret)
		- [x] private (personal private fork)

The remaining private github.com case has high confidence that it works given the self-hosted private case (it's basically the same). There are other edge cases worth considering, like if the taxonomy repo itself is ssh based but the `qna` repos are git based, the code currently doesn't support this but it should be a simple addition to add this support (probably enough to get rid of the conditional on ssh vs username/pat, but there may be other issues with this). 

Note that the git clone component is removed entirely, this is because we don't want to be passing token/credentials as input/outputs around, and if we kept this as a separate component there would be a lot of repeated logic in both the clone and sdg_op components. To avoid this we just manage the auth, clone, and sdg generate in the same component. 

## How Has This Been Tested?
Ran the ilab pipeline in a non-disconnected environment

I haven't tested it with tolerations & node selectors yet, but the rest works.

I've tested the sdg taxonomy repo auth against the cases mentioned above.

Log Outputs from the different cases outlined above: 

Self-Signed cases
[self-signed-https-private-repo.log](https://github.com/user-attachments/files/19035218/logs-self-signed-private.log)
[self-signed-https-public-repo](https://github.com/user-attachments/files/19035220/logs-self-signed-public.log)

Github.com Cases (well-known)
[github-https-public-repo.log](https://github.com/user-attachments/files/19035207/logs-github-https-public.log)
[github-ssh-private-repo.log](https://github.com/user-attachments/files/19035215/logs-github-ssh-private.log)
[github-ssh-public-with-ssh-key.log](https://github.com/user-attachments/files/19035217/logs-github-ssh-public-with-key.log)

## Merge criteria:


eval/final.py
```diff
@@ -341,9 +341,49 @@ def find_node_dataset_directories(base_dir: str):
...
+
+        if response.status_code == 200:
+            secret_data = response.json().get("data", {})
+            return [base64.b64decode(secret_data[key]).decode() for key in keys]
+        else:
+            raise RuntimeError(
+                f"Error fetching secret: {response.status_code} {response.text}"
+            )
+
+    if judge_secret_name is None:
```
HumairAK, line 377: side note on this, we keep the previous approach for 2 reasons: 

1. we dont' want to break standalone.py, and this code is leveraged there so we need to maintain backwards compatibility 
2. we will need this again when we use the sdk to mount the secrets, and we'll get rid of the new additional calls to fetch_secret

eval/final.py
```diff
@@ -341,9 +341,49 @@ def find_node_dataset_directories(base_dir: str):
...
+            raise RuntimeError("Error reading service account token")
+
+        headers = {"Authorization": f"Bearer {token}", "Accept": "application/json"}
+        verify_tls = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+        url = f"{K8S_API_SERVER}/api/v1/namespaces/{namespace}/secrets/{secret_name}"
+        response = requests.get(url, headers=headers, verify=verify_tls)
+
+        if response.status_code == 200:
+            secret_data = response.json().get("data", {})
+            return [base64.b64decode(secret_data[key]).decode() for key in keys]
```
mprahl, line 371: The call to `os.getenv("JUDGE_API_KEY", "")` below implies it is optional but this code would produce a KeyError exception if the key isn't present.

eval/final.py
```diff
@@ -341,9 +341,49 @@ def find_node_dataset_directories(base_dir: str):
...
+        try:
+            with open(TOKEN_PATH, "r") as f:
+                token = f.read().strip()
+        except FileNotFoundError:
+            raise RuntimeError("Error reading service account token")
+
+        headers = {"Authorization": f"Bearer {token}", "Accept": "application/json"}
+        verify_tls = "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+        url = f"{K8S_API_SERVER}/api/v1/namespaces/{namespace}/secrets/{secret_name}"
+        response = requests.get(url, headers=headers, verify=verify_tls)
```
mprahl, line 367: Could you set a timeout? Python requests will hang forever in some cases if there isn't a timeout set.

pipeline.py
```diff
@@ -392,12 +392,6 @@ def ilab_pipeline(
     run_mt_bench_task.set_accelerator_limit(1)
     run_mt_bench_task.set_caching_options(False)
     run_mt_bench_task.after(training_phase_2)
-    use_config_map_as_env(
```
mprahl, line 385: Should this have been removed to maintain backwards compatibility?

sdg/components.py
```diff
@@ -91,19 +96,171 @@ def fetch_secret(secret_name, keys):
...
+        print("Configuring Git Credentials...")
+        # Parse the domain from the repository URL
+        parsed_url = urllib.parse.urlparse(repo_url)
+        git_server = parsed_url.netloc
+
+        # Set up Git credential helper
+        exec_cmd(["git", "config", "--global", "credential.helper", "store"], env=env)
+
+        # Save credentials dynamically for any Git server
+        git_credentials_path = "/data/.git-credentials"
```
mprahl, line 186: Could you store these using tempfile.NamedTemporaryFile? Not sure if /data is to a PVC but we likely don't want the Git credentials permanently stored there and using a context manager such as with tempfile.NamedTemporaryFile allows it to clean up even if there is an exception below.

sdg/components.py
```diff
@@ -91,19 +96,171 @@ def fetch_secret(secret_name, keys):
...
+            stdout=open(os.path.expanduser("~/.ssh/known_hosts"), "a"),
+        )
+
+        # A repo can rarely support anonymous ssh access (rare)
+        # So we conditionally check for the key.
+        if ssh_key:
+            print("Configuring SSH authentication for Git...")
+            # Write the SSH key to a temporary file
+            with tempfile.NamedTemporaryFile(
+                delete=False, mode="w", prefix="ssh_key_", dir=ssh_dir, suffix=".pem"
```
mprahl, line 224: The credentials should be cleaned up after use in my opinion.

sdg/components.py
HumairAK: ah yeah good catch, I did this for the ssh creds but forgot to do it here, will fix this

sdg/components.py
```diff
@@ -91,19 +96,171 @@ def fetch_secret(secret_name, keys):
...
+
+    # Username/PAT takes precedence over ssh
+    if username and token:
+        print("Configuring Git Credentials...")
+        # Parse the domain from the repository URL
+        parsed_url = urllib.parse.urlparse(repo_url)
+        git_server = parsed_url.netloc
+
+        # Set up Git credential helper
+        exec_cmd(["git", "config", "--global", "credential.helper", "store"], env=env)
```
mprahl, line 183: Optional: Personally, rather than setting the configuration globally, I would just keep track of the args to pass to the Git command with something like `--config-env=<name>=<envvar>` (if you're worried about an exception showing the command with credentials) or `-c` (if you are able to not log the credentials):
https://git-scm.com/docs/git/2.34.0#Documentation/git.txt---config-envltnamegtltenvvargt
https://git-scm.com/docs/git/2.34.0#Documentation/git.txt--cltnamegtltvaluegt

HumairAK: the reason we do it this way is because ilab sdg generate code will also do some git clones, and they don't allow env passing, so this way they can implicitly inherit those configurations

mprahl: Ah, okay. That makes sense.

pipeline.py
HumairAK: Not sure I'm following, backwards compatibility for what?

mprahl: The code under `if judge_secret_name is None:` relies on an environment variable of `JUDGE_ENDPOINT` and `JUDGE_NAME` does it not?

HumairAK: oh I see what you mean, the backwards compatibility comment I left in the PR description refers to only standalone.py, which will utilize the same component code for sdg/mt_eval/final_eval but will mount the configmaps/secrets as env vars, for example for sdg this is done [here](https://github.com/opendatahub-io/ilab-on-ocp/blob/d9678da2a8939c18d0f2a363789559cbd3ca3534/standalone/standalone.py#L1291), we want to maintain compatibility with standalone.py

from the pipeline's perspective, you provide a secret name, and we will use it, how we use it is an implementation detail

eval/final.py
```diff
@@ -341,9 +341,54 @@ def find_node_dataset_directories(base_dir: str):
...
+                f"Error fetching secret: {response.status_code} {response.text}"
+            )
+
+    if judge_secret_name is None:
+        judge_api_key = os.getenv("JUDGE_API_KEY", "")
+        judge_model_name = os.getenv("JUDGE_NAME")
+        judge_endpoint = os.getenv("JUDGE_ENDPOINT")
+    else:
+        print("Eval Judge secret specified, fetching...")
+        judge_api_key, judge_model_name, judge_endpoint = fetch_secret(
```
mprahl, line 388: This will get a ValueError if keys are missing in the return value. `fetch_secret` should have an else to the `if key in secret_data:` to set an empty string.

Let's address this in a followup PR.

mprahl (approved):

PR #280, opened 2025-03-07
Add support for PDFs in SDG

mprahl:
## Description

Taxonomy repos with PDFs require Docling in RHEL AI 1.3, which requires the Mixtral tokenizer. In RHEL AI 1.4, Docling will always be used so this is also required for that upgrade.

Relates:
https://issues.redhat.com/browse/RHOAIENG-21302

## How Has This Been Tested?

1. Using the latest ODH main branch
2. Forked https://github.com/juliadenham/Summit_knowledge.git and converted swifties.md to swifties.pdf
3. Forked https://github.com/instructlab/taxonomy and had it point to the knowledge fork and use swifties.pdf
4. Ran a pipeline pointing to the taxonomy fork
5. Confirmed the SDG step worked and didn't download Docling from HuggingFace

## Merge criteria:

### DIFF
--- a/sdg/components.py
+++ b/sdg/components.py
@@ -53,6 +53,7 @@ def git_clone_op(
 def sdg_op(
     num_instructions_to_generate: int,
     pipeline: str,
+    tokenizer_model: dsl.Input[dsl.Model],
     repo_branch: Optional[str],
     repo_pr: Optional[int],
     taxonomy_path: str = "/data/taxonomy",
@@ -64,6 +65,7 @@ def sdg_op(
 ):
     import base64
     import os
+    import os.path
     import re
     import shutil
     import subprocess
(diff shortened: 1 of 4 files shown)


sdg/components.py
```diff
@@ -404,7 +429,7 @@ def set_precomputed_skills_data_ratio(sampling_size: float, skills_recipe: str):
                         output_dir=sdg_path,
                         taxonomy=taxonomy_path,
                         taxonomy_base=taxonomy_base,
-                        model_name=model,
+                        model_name="mixtral",
```
rimolive, line 432: Should this be hard-coded?

sdg/components.py
```diff
@@ -367,6 +378,9 @@ def set_precomputed_skills_data_ratio(sampling_size: float, skills_recipe: str):
                 ]
                 temp_pipeline_dir = os.path.join(temp_dir, "pipeline")
                 os.mkdir(temp_pipeline_dir)
+                temp_models_dir = os.path.join(temp_dir, "instructlab", "sdg", "models")
+                os.makedirs(temp_models_dir, exist_ok=True)
```
rimolive, line 382: Optional: It would be good to set proper permissions to avoid potential Permission Denied errors (assuming this path is stored in a PVC).

sdg/components.py
```diff
@@ -322,7 +331,7 @@ def get_git_host(repo_url):
             output_dir=sdg_path,
             taxonomy=taxonomy_path,
             taxonomy_base=taxonomy_base,
-            model_name=model,
+            model_name="mixtral",
```
rimolive, line 334: Should this be hard-coded?

rimolive (commented):
I added some comments.

sdg/components.py
mprahl: Yes, because the embedded tokenizer is Mixtral.

sdg/components.py
mprahl: This is coming from the `tempfile.TemporaryDirectory` context manager so it's just in a directory in `/tmp`, so the current user will have write access and we don't have to worry about other users/pods having access since the data is deleted after the context manager is done executing.

sdg/components.py
mprahl: Yes, because the embedded tokenizer is Mixtral and that is the only one that RHEL AI supports right now from my understanding. If we add support for more in the future, we'll need to make this a parameter and have RHEL AI distribute tokenizers in a better way (not included in an ~80GB container image). The latter is tracked in:
https://issues.redhat.com/browse/RHELAIRFE-10

gmfrasca (commented):
/lgtm

PR #295, opened 2025-03-28
Fix test model registry task & other enhancements

jgarciao:
- Delete last / of model_registry_endpoint to fix connection errors
- Allow sdg_num_workers = 1
- Set default train_memory_per_worker to 56Gi

Tested successfully in a pipeline run (more details in this [slack thread](https://redhat-internal.slack.com/archives/C062MMU059S/p1743172827907329?thread_ts=1743170418.713919&cid=C062MMU059S))

### DIFF
--- a/utils/components.py
+++ b/utils/components.py
@@ -394,7 +394,7 @@ def test_sdg_params(sdg_batch_size: int, sdg_num_workers: int):
     if (
         sdg_batch_size < 1
         or sdg_batch_size > 4096
-        or sdg_num_workers < 2
+        or sdg_num_workers < 1
         or sdg_num_workers > 10
     ):
         print(
--- a/utils/components.py
+++ b/utils/components.py
@@ -444,6 +444,7 @@ def test_model_registry(
 
     try:
         # Extract the port out of the URL because the ModelRegistry client expects those as separate arguments.
+        model_registry_endpoint = model_registry_endpoint.rstrip("/")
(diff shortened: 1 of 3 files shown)


utils/components.py
```diff
@@ -394,7 +394,7 @@ def test_sdg_params(sdg_batch_size: int, sdg_num_workers: int):
     if (
         sdg_batch_size < 1
         or sdg_batch_size > 4096
-        or sdg_num_workers < 2
+        or sdg_num_workers < 1
```
mprahl, line 397: Unfortunately, we can't allow this because this causes InstructLab to disable batching and then sends each completion request in a separate HTTP call.

mprahl: Oh I see. This is intentional based on the docstring. What's the usecase for disabling batching?

jgarciao: The idea is to reduce the load on the teacher model to avoid SDG failing due to timeout errors (see https://issues.redhat.com/browse/RHOAIENG-20555#)

Also, the error message in test_sdg_params says `sdg_batch_size must be a value between 1-4096`

mprahl: We were told by RHEL AI that we always want batching enabled which is why I added this check originally since they said an HTTP request per completion creates a high load on the inference endpoint.

jgarciao: I've modified the code to ensure that sdg_num_workers is a value between 2-10

mprahl (approved):

PR #304, opened 2025-04-16
update to use k8s auth/ssh secret types for sdg repo

HumairAK:
## Description
Support basic auth and ssh secret types. This is an api breaking change as we're no longer supporting the old secret keys for taxonomy secret.

Tested with:

Secrets of format:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: secret-ssh-auth
type: kubernetes.io/ssh-auth
data:

  ssh-privatekey: |
    ...
```

and:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: secret-basic-auth
type: kubernetes.io/basic-auth
stringData:
  username: ...
  password: ...
```

Tested against private repo via user/pass and ssh, as well as teacher model connections.

### DIFF
--- a/sdg/components.py
+++ b/sdg/components.py
@@ -43,7 +43,7 @@ def sdg_op(
 
     REQUEST_TIMEOUT = 30  # seconds
 
-    def fetch_secret(secret_name, keys, optional=False):
+    def fetch_secret(secret_name, optional=False):
         # Kubernetes API server inside the cluster
         K8S_API_SERVER = "https://kubernetes.default.svc"
         NAMESPACE_PATH = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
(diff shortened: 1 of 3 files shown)


sdg/components.py
```diff
@@ -132,11 +125,40 @@ def get_git_host(repo_url):
         ssh_key = os.getenv("GIT_SSH_KEY")
     else:
         print("SDG repo secret specified, fetching...")
-        username, token, ssh_key = fetch_secret(
-            taxonomy_repo_secret,
-            ["GIT_USERNAME", "GIT_TOKEN", "GIT_SSH_KEY"],
-            optional=taxonomy_repo_secret == "taxonomy-repo-secret",
-        )
+        # When the user opts not provide a taxonomy repo
```
mprahl, line 128:
```suggestion
        # When the user opts not provide a taxonomy repo secret
```

sdg/components.py
```diff
@@ -273,9 +295,17 @@ def get_git_host(repo_url):
...
+        )
+        secret_data = secret.get("data", {})
+        api_key = (
+            base64.b64decode(secret_data["api_token"]).decode()
+            if "api_token" in secret_data
+            else ""
         )
+        model_name = base64.b64decode(secret_data["model_name"]).decode()
+        endpoint = base64.b64decode(secret_data["endpoint"]).decode()
         if endpoint is None or model_name is None:
```
mprahl, line 309: This will throw a KeyError if the user made a mistake on the secret. How about this?
```suggestion
        model_name = base64.b64decode(secret_data.get("model_name", "")).decode()
        endpoint = base64.b64decode(secret_data.get("endpoint", "")).decode()
        if not endpoint or not model_name:
```

mprahl (approved):
/approve
/lgtm

